PR #1, opened 2019-10-08
Velocity interface

fmauch:
This MR aims at adding a velocity interface.
Please note: This is currently not thoroughly tested and under development. Expect the velocity-based controllers to behave non-stable, especially the joint_trajectory controller.
Speed scaling is currently not supported by the interface, but could be added, once a stable joint_trajectory_controller could be added.
If you still want to go ahead and test this, please do so and let us know of any issues or even better help developing this feature.
In general, this will help getting the robot's response time down to two cycles ideally.


AndyZe:
I compared a JointGroupPositionController against a JointGroupVelocityController. Surprisingly, the delay for the JointGroupVelocityController was worse. This was with Ubuntu 18, regular kernel, UR3. Any hints on why? Do I need a real-time kernel to achieve the 2 control cycle delay?

position_control_annotated
velocity_control_annotated

fmauch:
Hi, thank you for your plots.

This is somewhat complicated, as there are a lot of different places where delays can come from.

If I understand your plots correctly, then there is the following chain:
 1. Command generation in some jogging implementation
 2. Topic communication between jogging and controller (and between jogging and plotting, which is not necessarily with the same time delay)
 3. Update cycle of the driver taking the latest command value
 4. Sending Command to robot
 5. Robot sets new setpoint (in speed domain, using speedj internally)
 6. Robot control is moving towards the specified target
 7. Current status is sent to ROS pc
 8. Updated joint values are sent to robot_state_publisher in next update cycle
 9. Updated link position is sent to tf

While many of those delays do not show up in your plots if the plots are actually using the timestamps, I think it is clear that "delay" is not so easy to specify.

What I meant with the statement

> In general, this will help getting the robot's response time down to two cycles ideally.

was the delay between 4. and 6.

Let me also add a plot to further elaborate what I mean:
image

We see three curves representing the "same" velocity of the last robot axis.
  * In red we see the data that is being sent to the robot in the controller's `write` function.
  * The robot's internal setpoint is depicted in teal
  * The actual joint speed is shown in magenta

What we can see is:
  * The robot is actually moving the commanded velocity into its internal setpoint in one cycle (in terms of starting a motion, leaving the 0-value)
  * The robot doesn't follow the setpoint with the same "aggressiveness" as commanded. This is basically (from my understanding) due to the actual speedj interpolation.
  * The robot's actual motion is further delayed by the internal controller, so this will result in a further delay viewn from tf perspective

There's one more factor to influence this, which is the `speedj` parametrization inside the URScript (`resources/servoj.urscript`, I haven't renamed the script on that branch, yet). You can fiddle around with the acceleration and timing parameters, however from my try-and-error changing these from the value currently being present made things worst in most cases.

At last, let me comment on your real-time-Kernel question:
Using a real-time kernel might improve timings between 2 and 4, as the controller thread is running with real-time scheduling, as well as the actual communication with the robot and thereby improving 4 and 7. This does not mean, that it will be faster, it will be more reliable that your update cycle is called in a 500Hz frequency. On a non-real-time kernel we had dime deltas between update cycles much more spread all over the place.

I hope, this answers your questions, if not, let me know.

gavanderhoorn:
@fmauch: could you please rebase `velocity_interface` on current `master`? There are quite a few commits missing from this branch.

fmauch:
I just rebased it to the current master.

I'm not happy with that, as I couldn't get the velocity based trajectory controller running on the real hardware, yet. I could, however, merge this with the joint group velocity controller only. This way, people could use this for velocity streaming or for using their own velocity-based controllers.

gavanderhoorn:
> as I couldn't get the velocity based trajectory controller running on the real hardware, yet.

As in: you haven't had a chance to test it yet, or it really doesn't work yet?

fmauch:
> > as I couldn't get the velocity based trajectory controller running on the real hardware, yet.
> 
> As in: you haven't had a chance to test it yet, or it really doesn't work yet?

As in I tried it and it was causing protective stops in many situations. I've been playing around (not optimizing) the controller's pid parameters but without anything coming close to real usability. I also tried implementing joint limits which actually made things worse...

However, I haven't put too much time into it.

gavanderhoorn:
Hm.

Perhaps releasing this for *just* the `JointGroupVelocityController` would be an idea.

But it would not prevent anyone from configuring things to use a velocity-based `JointTrajectoryController`, which will lead to support questions.

alfaDoG1993:
I tested this branch on a real UR3 with DualShock 4 V2 controller. Everything seems to be working fine with the `joint_group_vel_controller`.

sinback:
Hi all. I tried adapting an application with the UR10 to use this URScript portion driver during the past week. We actually don't use ROS itself, so I can't comment about how well the driver integrates with ROS, but I have some observations about the URScript part of the driver which were especially relevant to our application and might be relevant for you too. Hopefully the discussion can be of value to you.

We are migrating from a control scheme which worked well with UR's CB3 series, which simply sends speedj or servoj commands to the robot, on the recommended 8ms control loop period. This worked very poorly with the e-series robots, for reasons we're still not sure about, so we decided to start using the ROS style of doing things instead on the UR side. However my understanding is that doing motion planning in step with the 8ms control loop is not quite the ROS way of doing things (hence eg. the servoj interpolation feature, which we don't end up leveraging). We noticed though that when using the speedj thread for control, as a drop-in replacement to sending speedj commands every 8ms, the robot had a lot of quite bad issues with motion stutter, etc. which we'd never observed with just using speedjs, which initially made us think we'd need to massively slow down our motion speed to get smooth/estopless motion behavior.

However after staring at the code and hacking on it some I finally tried two things: removing the keepalive signalling from the robot back to the control PC (I imagine this would be really annoying for the ROS environment, but it does save the robot time doing socket transactions), and also removing the calls to sync() in the motion control threads (something it might make more sense to try in the ROS environment). I noticed that the performance of the robot suddenly became nearly comparable to how it used to be, without us having to update our motion planning algorithms. It was really a big improvement for us, although I haven't quite wrapped my head around why yet. However I have a hypothesis about it. Here's the function description from the URScript 3.2 manual on speedj:

**speedj(qd, a, t)**
**Joint speed**
Accelerate linearly in joint space and continue with constant joint
speed. **The time t is optional; if provided the function will return after
time t, regardless of the target speed has been reached.** If the time t is
not provided, the function will return when the target speed is reached.
Parameters
qd: joint speeds [rad/s]
a: joint acceleration [rad/sˆ2] (of leading axis)
t: time [s] before the function returns (optional)

I don't think servoj works this way, although I'm not sure how long it takes for it to return. However given how speedj is documented as working, I think maybe what is happening is that since the speedj takes nearly an entire steptime amount of time to complete (you give it 0.9*steptime in this PR), having a call to sync() right afterwards makes it so that the speedj can actually sometimes only run at every 2 steptime periods? If the rest of the work in the main thread takes more than 800us for the CB3 series or 200us for the e-series, the work of one run of this thread will actually just take longer than the steptime period we have to get everything done by. I'm really not sure about anything regarding the hardware or OS the UR control box runs but it seems to me like like that could probably happen often, especially in the case of the socket transactions with the PC client taking a long period of time? (It sounds like it could be related to the issue @AndyZe mentioned above too?) Once I removed the calls to sync() I could successfully adjust the time speedj took to be fully a steptime period rather than the 0.9*steptime period you have specified here, which seemed like a beneficial move to me. I think doing so means that the speedj control is no longer in lock-step with the steptime-length control timeslots divided up on the robot, but it still allows for good closed-loop motion planning with our application.

I'm still not totally convinced of the reasoning above, especially because I haven't profiled the socket transactions to see if they actually would end up taking longer than the time window they have to complete and still allow the speedj to finish within its timeslot, but I think it might be right.

I've so far only seen positive consequences of removing the calls to sync() yet but it certainly felt like a weird choice to make, since it was also present in the servoj control thread for this URScript driver and it seems like it's been around for quite a while in it -- Do you foresee negative consequences to this choice? If not it might be worth testing out for yourself. If so, I'd be really curious to hear about them since it might be helpful to me too! :)

fmauch:
Good timing! I was just about to tackle this again to finally merge it.

From my understanding of UR's threading model the keepalive signal (I guess, you are talking about [this line](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/velocity_interface/ur_robot_driver/resources/servoj.urscript#L110)) could indeed interfering with this. Moving it out of the critical section might help, so the critical section only contains updating the setpoint (or joint speeds in this case).

To be honest, I don't quite remember why we had the sync in there in the first place and we had the `0.9 * steptime` (And the comment indicates that I didn't know for sure in the past, either).

However, from my understanding of UR's threading model it does make sense to return from the `speedj` command quite early. When execution of the speedj command starts in the middle of one cycle slot and we let it run for a whole slot time, it would in fact run over 2 time slots. I might be completely off here, though. I'll try to run a couple of tests again and talk to some people and then get back to this. Thanks anyway for picking this up @sinback

fmauch:
I sat down this afternoon and now I think, we've got a proper solution. @sinback You were absolutely right, that `sync()` was nonsense. I believe, this was a leftover from the time where I understood the documentation in such a way that if we don't pass a time, it will keep the set speed as long as no other command is sent. With giving it the steptime as time it seems to receive a new setpoint before actually stopping. Would you agree with the current implementation?

With this, the velocity-based trajectory controller is also working again.

If no further input is coming, I will start cleaning up this branch (And the controller files) to make this ready to merge. I'll remove the experimental flag from the top, as now it is actually working as expected.

sinback:
I would guess that the robot actually does keep moving at speedj's set speed until it's told not to, but I don't see how a call to sync() would change anything about that, and in this situation, nothing too bad would come of that in this situation if the ROS driver crashed, because the URScript program exits if its keepalive signal goes away (and when a URScript program concludes the robot does stop moving).

I would put the reading of qd in the speedThread back in a critical section since qd is set in the main thread and it's good to ensure that the speed thread isn't reading stale/partially-stale values.

It's good to hear my poking helped you out :)

fmauch:
> I would guess that the robot actually does keep moving at speedj's set speed until it's told not to, 

Apparently, it doesn't

> but I don't see how a call to sync() would change anything about that, 

The sync is needed, as the thread otherwise idles around after the speedj has finished (if given a shorter time)

> I would put the reading of qd in the speedThread back in a critical section since qd is set in the main thread and it's good to ensure that the speed thread isn't reading stale/partially-stale values.

Well, critical sections are not like semaphores, I got that wrong in the past, as well. They just rather say that commands inside a critical section get executed one after another without any other thread interrupting. As there was only one statement inside the critical section, it was basically useless.

sinback:
Thanks for the information!

ur_robot_driver/config/ur10_controllers.yaml
```diff
@@ -60,3 +60,58 @@ pos_traj_controller:
    stop_trajectory_duration: 0.5
    state_publish_rate: 125
    action_monitor_rate: 10
+
+vel_based_pos_traj_controller:
+   type: velocity_controllers/JointTrajectoryController
+   joints:
```
gavanderhoorn: Could this (and other files) not use the `*robot_joints` 'trick'?

fmauch: This is what I meant by "I will start cleaning up this branch (And the controller files)" above :-)

ipa-lth:
This rebased merge causes some trouble on our setup.
It seems that some thread on the ur-controller is not stopped properly, before executing a following motion. Thus, as soon as the "externalControl" is finished by the "handback control" service call, the ur continues to the next motion (in my case a ptp motion) an fails.

To test this you have to create a ur program containing the "ExternalControl" followed by a "MoveJ".
Launch the ros (control) driver and the ur program, and manually trigger the "hand-back-control" service.
I tested it on ursim3.7.4 and ursim3.12 with the same results.

1579781424874 remmina-2020-1-23-13:25:41 518376

I merged fmauch:velocity_interface to the master end of last year on [my own fork](https://github.com/ipa-lth/Universal_Robots_ROS_Driver/tree/velocity_interface). This can be considered a alpha "unclean version". Here there is no such threading problem and an endless loop switching between "extControl" and "ur motions" works as expected.

jfelip:
Hi, I'm observing the same problem on a real robot (UR5 CB3) and only using the ExternalControl without any other instruction in the program. The arm accepts the commands from the ROS driver and moves as expected until eventually stops showing the same error "Another thread is already controlling the robot". Our Polyscope version is 3.9.1.

fmauch:
@ipa-lth There was a join missing at the end of the program. While I was able to reproduce the error you mentioned, with eedf1fb applied this is gone.

@jfelip Do you also call the `hand_back_control` service or does it pop up all of a sudden? Are you doing any controller switch? What do the log messages say in such a situation?

gavanderhoorn:
Looks like d7d6ad5b includes a file called `t` which should not be included ([this one](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/1/commits/d7d6ad5b330d2e2f5d739dfc1c333a9161255bdd#diff-e358efa489f58062f10dd7316b65649e)).

fmauch:
> Looks like [d7d6ad5](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/commit/d7d6ad5b330d2e2f5d739dfc1c333a9161255bdd) includes a file called `t` which should not be included ([this one](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/1/commits/d7d6ad5b330d2e2f5d739dfc1c333a9161255bdd#diff-e358efa489f58062f10dd7316b65649e)).

whoopsie. Thanks.

ipa-lth:
> @ipa-lth There was a join missing at the end of the program. While I was able to reproduce the error you mentioned, with [eedf1fb](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/commit/eedf1fb1a53271b06e6c3cbfd1c7a7d75b4db5b8) applied this is gone.

With the join statement, the control urcap never ends here. When I call the hand_back_control service it says "True; Deactivated control", but the ur programm does not continue... Am I missing something?

fmauch:
> > @ipa-lth There was a join missing at the end of the program. While I was able to reproduce the error you mentioned, with [eedf1fb](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/commit/eedf1fb1a53271b06e6c3cbfd1c7a7d75b4db5b8) applied this is gone.
> 
> With the join statement, the control urcap never ends here. When I call the hand_back_control service it says "True; Deactivated control", but the ur programm does not continue... Am I missing something?

Actually, the service's return is only a driver internal. If the program on the control box hangs this won't be noticed here.

> 
> Edit: I am pretty sure, that the thread_move hangs and cannot be joined. Unfortunately, I don't know why.. and not with the version before...

Did you check with df1daee applied, as well? This should explicitly set the `control-mode` to `MODE_STOPPED` before joining the thread. This should bring the move thread (no matter whether position or velocity based) to an end. You can see the `control_mode` on the TP's run view. It should be 2 in velocity mode, 1 in position mode, 0 with no controller active and -2 when you handed back the control.

ipa-lth:
> Did you check with [df1daee](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/commit/df1daee172c33a5b975568aa23b4ee72ed5f399a) applied, as well?
I did. But the "control_mode = MODE_STOPPED" is commented out.

With this statement it works for me [lth-origin](https://github.com/ipa-lth/Universal_Robots_ROS_Driver/commit/a0402be2f999bf8bce2659360076c058c25a94c7)




> You can see the `control_mode` on the TP's run view. It should be 2 in velocity mode, 1 in position mode, 0 with no controller active and -2 when you handed back the control.

At a first glimpse TP look ok.

jfelip:
> @jfelip Do you also call the `hand_back_control` service or does it pop up all of a sudden? Are you doing any controller switch? What do the log messages say in such a situation?

No, I'm just streaming velocities to the joint_group_vel_controller and the error pops up all of the sudden. Do you mean the ROS log messages? If not, are there other log locations I should check?

fmauch:
@jfelip Sorry, that was not clear. I meant log messages on the robot's TP.

gavanderhoorn:
@fmauch: I don't remember, but do the log messages on the TP get prefixed by the name of the program which emits them?

If not: that could be a good thing to add.

fmauch:
> @fmauch: I don't remember, but do the log messages on the TP get prefixed by the name of the program which emits them?
> 
> If not: that could be a good thing to add.

True. I'll open a separate enhancement issue for this.

ipa-lth:
In combination with #93 this PR works fine with our "longterm" (~2h) test.

Issue #14, opened 2019-10-21
Finish headless mode


fmauch:
Currently, we only have a light headless mode, that starts the program initially. For reacting on errors and restarting the program after stops, we need access to more interactive functionality.

   * [x] Start booted robot (Initial enable) -> Dashboard
   * [x] Power off robot -> Dashboard
   * [x] Handle Protective Stop -> Dashboard
   * [x] Handle EM-Stop -> Dashboard
   * [x] Handle Fault -> Dashboard
   * [x] Handle violation -> Dashboard
   * [x] Publish robot status from RTDE
   * [x] Print robot mode and safety mode to logging
   * [x] Offer Action to start/restart/recover robot
   * ~~-> How to switch on the robot without a TP connected?~~ Not part of the driver

On top of that a state machine handling robot recoveries might be useful, currently, one has to perform multiple service calls to reenable the robot after an EM stop. This could, however, also be fine.

fmauch:
On RTDE `robot_mode` is specified as integer:

Mode|Description
-- | --
-1 | ROBOT_MODE_NO_CONTROLLER
0 | ROBOT_MODE_DISCONNECTED
1 | ROBOT_MODE_CONFIRM_SAFETY
2 | ROBOT_MODE_BOOTING
3 | ROBOT_MODE_POWER_OFF
4 | ROBOT_MODE_POWER_ON
5 | ROBOT_MODE_IDLE
6 | ROBOT_MODE_BACKDRIVE
7 | ROBOT_MODE_RUNNING
8 | ROBOT_MODE_UPDATING_FIRMWARE

while safety status information is available using status bits in the field `safety_status_bits` or an integer representation on `safety_mode`:


  | Value | CB3 | e-Series
-- | -- | -- | --
SAFETY_MODE_UNDEFINED_SAFETY_MODE | 11 | X | X
SAFETY_MODE_VALIDATE_JOINT_ID | 10 |   | X
SAFETY_MODE_FAULT | 9 | X | X
SAFETY_MODE_VIOLATION | 8 | X | X
SAFETY_MODE_ROBOT_EMERGENCY_STOP | 7 | X | X
SAFETY_MODE_SYSTEM_EMERGENCY_STOP | 6 | X | X
SAFETY_MODE_SAFEGUARD_STOP | 5 | X | X
SAFETY_MODE_RECOVERY | 4 | X | X
[...]

PR #16, opened 2019-10-23
Add a dashboard client to the driver

fmauch:
This MR adds a dashboard client forwarded as ROS service calls. This will be needed for providing full headless mode (#14) as this enables the user to restart and enable the robot after protective stops, EM-stops, etc., as well as initial bootup.

Currently, only parts of the actual interface is wrapped into service calls, as I did not implement custom services, yet, but used std_srvs/Trigger where applicable.

It might be reasonable to create additional srvs inside a new package `ur_dashboard_msgs`or as part of `ur_msgs`. Any thoughts on what would be better? @gavanderhoorn


ur_robot_driver/src/ur/dashboard_client.cpp
```diff
@@ -0,0 +1,347 @@
...
+  {
+    return false;
+  }
+  return true;
+}
+
+bool DashboardClient::clearOperationalMode()
+{
+  std::string answer = sendAndReceive("clear operational mode\n");
+  if (answer != "operational mode is no longer controlled by Dashboard Server")
```
gavanderhoorn: Should the condition here be inverted (ie: `==`)? I find the documentation for this somewhat confusing.

Perhaps a comment here would be good?

ur_robot_driver/include/ur_robot_driver/ros/dashboard_client.h
```diff
@@ -0,0 +1,109 @@
...
+
+#include <ros/ros.h>
+#include <std_srvs/Trigger.h>
+#include <std_srvs/Empty.h>
+
+#include <ur_robot_driver/ur/dashboard_client.h>
+
+namespace ur_driver
+{
+const int DASHBOARD_SERVER_IP = 29999;
```
gavanderhoorn: This seems to just be a constant for the port, not an ip.

ur_robot_driver/src/ros/dashboard_client.cpp
```diff
@@ -0,0 +1,130 @@
...
+  pause_service_ = nh_.advertiseService("pause", &DashboardClientROS::pause, this);
+  play_service_ = nh_.advertiseService("play", &DashboardClientROS::play, this);
+  power_off_service_ = nh_.advertiseService("power_off", &DashboardClientROS::powerOff, this);
+  power_on_service_ = nh_.advertiseService("power_on", &DashboardClientROS::powerOn, this);
+  quit_service_ = nh_.advertiseService("quit", &DashboardClientROS::quit, this);
+  restart_safety_service_ = nh_.advertiseService("restart_safety", &DashboardClientROS::restartSafety, this);
+  shutdown_service_ = nh_.advertiseService("shutdown", &DashboardClientROS::shutdown, this);
+  stop_service_ = nh_.advertiseService("stop", &DashboardClientROS::stop, this);
+  unlock_protective_stop_service_ =
+      nh_.advertiseService("unlock_protective_stop", &DashboardClientROS::unlockProtectiveStop, this);
```
gavanderhoorn: Could consider using lambdas here. Most service callbacks are single line 'forwards' to the `client_` member.

IIRC, lambdas should be usable as service server callbacks.

ur_robot_driver/src/ur/dashboard_client.cpp
```diff
@@ -0,0 +1,347 @@
...
+  switch (mode)
+  {
+    case OperationalMode::MANUAL:
+      mode_string = "manual";
+      break;
+    case OperationalMode::AUTOMATIC:
+      mode_string = "automatic";
+      break;
+    default:
+      throw std::runtime_error("Enum value not known.");
```
gavanderhoorn: Consider `throw`ing `std::invalid_argument`?

This also appears to be the only place where an exception is thrown, instead of returning `false`. Is that intentional or a deviation?

ur_robot_driver/src/ur/dashboard_client.cpp
```diff
@@ -0,0 +1,347 @@
...
+
+std::string DashboardClient::programState()
+{
+  std::string answer = sendAndReceive("programState\n");
+  return answer;
+}
+
+std::string DashboardClient::polyscopeVersion()
+{
+  std::string answer = sendAndReceive("PolyscpoeVersion\n");
```
gavanderhoorn:
```suggestion
  std::string answer = sendAndReceive("PolyscopeVersion\n");
```

gavanderhoorn:
> It might be reasonable to create additional srvs inside a new package `ur_dashboard_msgs`or as part of `ur_msgs`.

We can host them in `ur_msgs`, although they are rather specific to the dashboard server. It would perhaps be good to have some examples of custom service (or: action) files before we decide.

ur_robot_driver/include/ur_robot_driver/ros/dashboard_client.h
fmauch: yep, found that, as well :)

ur_robot_driver/src/ros/dashboard_client.cpp
fmauch: I changed this in the last commit to make the actual client very slim as interpreting the server's answer at such a low level seemed wrong to me. This makes this kind of obsolete, I guess.

ur_robot_driver/src/ur/dashboard_client.cpp
fmauch: This also got removed in the latest refactoring. This inconsistency was one of the reasons to refactor this.

fmauch:
> > It might be reasonable to create additional srvs inside a new package `ur_dashboard_msgs`or as part of `ur_msgs`.
> 
> We can host them in `ur_msgs`, although they are rather specific to the dashboard server. It would perhaps be good to have some examples of custom service (or: action) files before we decide.

Then I'll start developing them inside this repo, we can then move them over if desired before merging this.

fmauch:
> Suggestion: use actions instead of services.

This is a valid point.

ur_robot_driver/src/ros/dashboard_client.cpp
gavanderhoorn: I'm not sure. There still appear to be many very simple service callbacks that could be changed to lambdas instead.

fmauch: You mean like that?

```c++
  // Service to release the brakes. If the robot is currently powered off, it will get powered on on the fly.
  brake_release_service_ = nh_.advertiseService<std_srvs::Trigger::Request, std_srvs::Trigger::Response>(
      "brake_release", [&](std_srvs::Trigger::Request& req, std_srvs::Trigger::Response& resp) {
        resp.message = client_.sendAndReceive("brake release\n");
        resp.success = std::regex_match(resp.message, std::regex("Brake releasing"));
        return true;
      });
```

Or do I miss a way of simplifying this even more?

gavanderhoorn: Yes, for instance. And seeing as it's basically the same pattern over and over, you may introduce a macro which takes a few arguments. That could reduce all of this to single line definitions for every service server.

gavanderhoorn:
> I'm currently not completely happy with this, as I don't like using macros, but a "double" lambda seemed not to work.

Could use something like (untested):

```c++
inline auto create_service_server(ros::NodeHandle const& nh, std::string const topic, std::string const command, std::string expected)
{
  return nh.advertiseService<std_srvs::Trigger::Request, std_srvs::Trigger::Response>(
    topic, [&](std_srvs::Trigger::Request& req, std_srvs::Trigger::Response& resp) {
      resp.message = client_.sendAndReceive(command);
      resp.success = std::regex_match(resp.message, std::regex(expected));
      return true;
  });
}
```

avoids the macro, is type-safe, etc.

fmauch:
I think, I tried something similar and it resulted in a runtime error. Should have checked this earlier... 

Of course, because of the [&] we leave a dangling ref to `command` and `expected`. I'll use this then :-)

fmauch:
@t-schnell please review this

ur_robot_driver/include/ur_robot_driver/ur/dashboard_client.h
```diff
@@ -0,0 +1,94 @@
...
+ * for the server's response.
+ *
+ * For documentation about the dashboard server, please see
+ *  - https://www.universal-robots.com/how-tos-and-faqs/how-to/ur-how-tos/dashboard-server-cb-series-port-29999-15690/
+ *  - https://www.universal-robots.com/how-tos-and-faqs/how-to/ur-how-tos/dashboard-server-e-series-port-29999-42728/
+ */
+class DashboardClient : public comm::TCPSocket
+{
+public:
+  DashboardClient(const std::string& host);
```
t-schnell: Missing Doxygen comment

ur_robot_driver/doc/ROS_INTERFACE.md
```diff
@@ -656,6 +760,124 @@ This is the actual driver node containing the ROS-Control stack. Interfaces docu
...
+    Dismiss a protective stop to continue robot movements. NOTE: It is the responsibility of the user to ensure the cause of the protective stop is resolved before calling this service.
+
+#### Parameters
+ * "**receive_timeout**" (Required)
+
+    Timeout after which a call to the dashboard server will be considered failure if no answer has been received.
+
+ * "**robot_ip**" (Required)
+
+    Please add description. See dashboard_client_node.cpp line number: 36
```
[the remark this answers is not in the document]
fmauch: Correct. fixed.

ur_robot_driver/include/ur_robot_driver/ros/dashboard_client.h
```diff
@@ -0,0 +1,122 @@
...
+namespace ur_driver
+{
+/*!
+ * \brief ROS wrapper for UR's dashboard server access. Many (not necessarily all) dashboard
+ * functionalities are wrapped into ROS services here.
+ */
+class DashboardClientROS
+{
+public:
+  DashboardClientROS(const ros::NodeHandle& nh, const std::string& robot_ip);
```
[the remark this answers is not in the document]
fmauch: fixed

ur_robot_driver/include/ur_robot_driver/ur/dashboard_client.h
fmauch: fixed

ur_robot_driver/src/ros/dashboard_client.cpp
```diff
@@ -0,0 +1,345 @@
...
+  // Service to show a popup on the UR Teach pendant.
+  popup_service_ = nh_.advertiseService<ur_dashboard_msgs::Popup::Request, ur_dashboard_msgs::Popup::Response>(
+      "popup", [&](ur_dashboard_msgs::Popup::Request& req, ur_dashboard_msgs::Popup::Response& resp) {
+        resp.answer = this->client_.sendAndReceive("popup " + req.message + "\n");
+        resp.success = std::regex_match(resp.answer, std::regex("showing popup"));
+
+        return true;
+      });
+
+  // Service the query the current program state
```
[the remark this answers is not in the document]
fmauch: fixed

ur_robot_driver/src/ros/dashboard_client.cpp
```diff
@@ -0,0 +1,345 @@
...
+            resp.success = std::regex_match(resp.answer, match, expected);
+            if (resp.success)
+            {
+              resp.state.state = match[1];
+              resp.program_name = match[2];
+            }
+            return true;
+          });
+
+  // Service the query the current safety mode
```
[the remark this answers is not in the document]
fmauch: fixed

ur_robot_driver/src/ros/dashboard_client.cpp
```diff
@@ -0,0 +1,345 @@
...
+  safety_mode_service_ = nh_.advertiseService("get_safety_mode", &DashboardClientROS::handleSafetyModeQuery, this);
+
+  // Service the query the current robot mode
```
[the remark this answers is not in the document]
fmauch: fixed

ur_robot_driver/src/ros/dashboard_client.cpp
```diff
@@ -0,0 +1,345 @@
...
+            return true;
+          });
+
+  // Service to reconnect to the dashboard server
+  reconnect_service_ = nh_.advertiseService<std_srvs::Trigger::Request, std_srvs::Trigger::Response>(
+      "connect", [&](std_srvs::Trigger::Request& req, std_srvs::Trigger::Response& resp) {
+        resp.success = connect();
+        return true;
+      });
+  //
```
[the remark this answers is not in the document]
fmauch: fixed

ur_robot_driver/src/ur/dashboard_client.cpp
```diff
@@ -0,0 +1,110 @@
...
+    return false;
+  }
+  bool ret_val = false;
+  if (TCPSocket::setup(host_, port_))
+  {
+    LOG_INFO("%s", read().c_str());
+    ret_val = true;
+  }
+  return ret_val;
+  ;
```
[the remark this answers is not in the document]
fmauch: fixed

Issue #6, opened 2019-10-16
Failed to read from stream, Pipeline producer overflowed!


Briancbn:
# Summary
After installing the externalcontrol-1.0.urcap and use my remote pc ip address as the remote host address, I launch the ur3e_bringup.launch file but I keep getting the error **Pipeline producer overflowed! <RTDE Data Pipeline>**, Communication error, what other setting steps did I miss?

# Versions
 - ROS Driver version:
 - Affected Robot Software Version(s): PolyScope 5.4
 - Affected Robot Hardware Version(s): ur3e
 - Robot Serial Number:
 - UR+ product(s) installed: ur3e
 - URCaps Software version(s): externalcontrol-1.0.urcap

# Issue details
![error_msg](https://user-images.githubusercontent.com/15899525/66914061-9505af00-f048-11e9-87c3-7dc902f19ea0.png)

## Use Case and Setup
Just testing the ur3e driver usability currently

## Steps to Reproduce
See summary

## Expected Behavior
I didn't expect there to be any error when running the ur3e_bringup program

Abbyls:
This error also happened to me, but not every time. And it did not affect my Moveit controlling.

fmauch:
Thanks for this clarification. I'll prepare a fix that should prevent this from happening.

Briancbn:
We managed to temporarily resolved the problem by switching to another PC. It happens all the time on the original PC I tested the code with, but the new PC seems to be ok for now. (I am not sure this helps)

fmauch:
I still think, that this is a bug. Which is why I asked whether the errors eventually stop and the driver is running normally. I'll provide a fix, anyway.

fmauch:
@Briancbn, @Abbyls could you please try to reproduce the issue using #18 applied?

fmauch:
Should be fixed in #18. Closing this. If it appears to be not fixed, please reopen.

ljden:
I'm getting this error while running the current stable (master) branch. Would you like me to open a new issue or reopen this one?

fmauch:
> I'm getting this error while running the current stable (master) branch. Would you like me to open a new issue?

Please open a new one describing your specific problem. If we realize that this is really the same problem as here, we can merge them later ( define the new issue a duplicate).

ljden:
> > I'm getting this error while running the current stable (master) branch. Would you like me to open a new issue?
> 
[...]

PR #18, opened 2019-10-28
Fix init timing

fmauch:
This MR aims at fixing startup timing issues like
 * Too much time goes by between starting the pipeline and starting robot control
 * Too much time goes by between the RTDE handshake and starting the pipeline

Also, we added more error handling (of cases that usually should not occur) to the RTDE handshake process.

This should solve #6

### DIFF
--- a/ur_robot_driver/include/ur_robot_driver/comm/pipeline.h
+++ b/ur_robot_driver/include/ur_robot_driver/comm/pipeline.h
@@ -189,6 +189,10 @@ public:
   {
   }
 
+  virtual void startProducer()
+  {
+  }
+
   /*!
    * \brief Reads packages from some source and produces corresponding objects.
    *
@@ -268,6 +272,13 @@ public:
     stop();
   }
 
+  void init()
+  {
+    producer_.setupProducer();
+    if (consumer_ != nullptr)
+      consumer_->setupConsumer();
+  }
+
   /*!
    * \brief Starts the producer and, if existing, the consumer in new threads.
    */
@@ -277,7 +288,7 @@ public:
       return;
 
     running_ = true;
-    producer_.setupProducer();
+    producer_.startProducer();
     pThread_ = std::thread(&Pipeline::runProducer, this);
     if (consumer_ != nullptr)
       cThread_ = std::thread(&Pipeline::runConsumer, this);
@@ -296,6 +307,7 @@ public:
 
     running_ = false;
 
+    producer_.stopProducer();
     if (pThread_.joinable())
     {
       pThread_.join();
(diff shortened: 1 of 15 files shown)


ur_robot_driver/src/rtde/rtde_client.cpp
```diff
@@ -47,14 +47,38 @@ RTDEClient::RTDEClient(std::string robot_ip, comm::INotifier& notifier, const st
...
+  pipeline_.init();
   pipeline_.run();
+
+  uint16_t protocol_version = negotiateProtocolVersion();
+  parser_.setProtocolVersion(protocol_version);
+
+  queryURControlVersion();
+  if (urcontrol_version_.major < 5)
+  {
+    max_frequency_ = CB3_MAX_FREQUENCY;
```
gavanderhoorn: Could be nice to notify users of this. Or at least add a `ROS_DEBUG..(..)`.

fmauch: It actually is still there. However, it got refactored into the `setOutputs()` function which is called directly afterwards.

Do you think, that would be nice to do, nonetheless? Output would then be something like
```
[ DEBUG] [1572871771.209652246]: Setting RTDE communication frequency to 125 Hz, as using a CB3 robot.
[ INFO] [1572871771.209652246]: Setting up RTDE communication with frequency 500.000000
```

fmauch:
@t-schnell please review

ur_robot_driver/include/ur_robot_driver/comm/producer.h
```diff
@@ -90,8 +99,10 @@ class URProducer : public IProducer<HeaderT>
    *
    * \returns Success of reading and parsing the package
    */
-  bool tryGet(std::vector<std::unique_ptr<URPackage<HeaderT>>>& products)
+  bool tryGet(std::vector<std::unique_ptr<URPackage<HeaderT>>>& products) override
   {
+    // TODO This function has become really ugly! That should be refactored!
+
```
t-schnell: Thinking about it more I don't think it has actually become worse, but some refactoring (or at least adding some comments) still seems like a good idea.

ur_robot_driver/include/ur_robot_driver/comm/pipeline.h
```diff
@@ -389,6 +401,8 @@ class Pipeline
     {
       if (!producer_.tryGet(products))
       {
+        producer_.teardownProducer();
+        running_ = false;
         break;
       }
```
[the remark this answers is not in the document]
fmauch: I think, tearing it down is actually reasonable here, as if this occurs, we want to reconnect anyway.

ur_robot_driver/include/ur_robot_driver/comm/producer.h
fmauch: Yes, it was worse in between... But I'd like to leave this comment there for the moment.

PR #30, opened 2019-11-07
Remove industrial_msgs dependency.

gavanderhoorn:
As per subject.

I believe this is a left-over from `ur_modern_driver` but is not actually used in the current implementation.


gavanderhoorn:
I would like to add infrastructure to the `hardware_interface` to be able to support the [industrial_robot_status_controller](http://wiki.ros.org/industrial_robot_status_controller), but that would still not need these messages here in this package.

fmauch:
Thank you for the contribution!

PR #31, opened 2019-11-07
Readme improvements (build instructions mostly)

gavanderhoorn:
As per subject.

Compile-tested in `ros:melodic`. Steps are copy-pastable.

I've added some empty lines so as to improve readability and group commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,7 +146,7 @@ to do so, as otherwise endeffector positions might be off in the magnitude of ce
 For this, there exists a helper script:
 
     $ roslaunch ur_calibration calibration_correction.launch \
-    robot_ip:=<robot_ip> target_filename:="${HOME}/my_robot_calibration.yaml"
+      robot_ip:=<robot_ip> target_filename:="${HOME}/my_robot_calibration.yaml"
 
 For the parameter `robot_ip` insert the IP address on which the ROS pc can reach the robot. As
 `target_filename` provide an absolute path where the result will be saved to.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -93,22 +93,27 @@ on information how to set this up.
...
-$ source devel/setup.bash
+sudo apt update -qq
+rosdep update
+rosdep install --from-path src --ignore-src -y
+
+# build the workspace
+catkin_make
+
+# activate the workspace (ie: source it)
+source devel/setup.bash
```
fmauch: I personally prefer having a `$` in front of each command, which makes it visually easier to understand when there are multi-line commands (at least in my opinion). If this is nothing usually being done, I'm fine to remove them.

gavanderhoorn: I understand why you say this, and I feel it does make line-continuation easier to spot, but I've answered enough questions on ROS Answers about "why does bash not execute my command which I just copy-pasted from a tutorial?" to not do this any more.


It's about 50/50 whether this is caused by incomplete copy-pastes, or because they copied line nrs and/or `$`-signs into the shell.

Inexperienced users will not realise they should not include the `$`, while experienced users will already know how to interpret these lists of commands. For those it will just make copy-pasting sets of commands more difficult.

As there is no way to toggle the `$`s, I've switched to removing them from snippets where the intent is for users to actually execute the commands verbatim (compared to, say, in examples which need the user to update the commands before they can execute them).

gavanderhoorn: If you'd prefer I could add the `$`s back though.

fmauch: That does make sense to me, yes. But then I'd prefer making this consistently through all code snippets in this repo's READMEs.

gavanderhoorn: > then I'd prefer making this consistently through all code snippets in this repo's READMEs

sure. But that seems a little out-of-scope for this PR, which was specifically about the build instructions in the main readme.

fmauch: Then I would vote for leaving in the `$`s for this PR and we tackle them in a separate commit.

gavanderhoorn:
Rebased and dropped the commit that removed the `$`.

fmauch:
Thanks for this contribution

PR #34, opened 2019-11-12
Support for combined robot hardware

asierfernandez:
We propose the minimum set of changes to use the driver together with [combined_robot_hw](https://github.com/ros-controls/ros_control/tree/melodic-devel/combined_robot_hw).

In our setup we use a haptic device with an UR robot using the `combined_robot_hw` and it works fine. This change doesn't break anything, and even if this is not the final solution, at least someone could use it as we do. 

For example, we see some limitations to how the `combined_robot_hw` together with the driver handles `read` operations. If I'm correct `read` is a blocking operation, and the `combined_robot_hw` executes this operation for all hardwares consecutively. Even if I haven't tested yet, if we combine multiple arms together with with some extra hardware it won't work properly.

There are many things to discuss about `combined_robot_hw`, and it would be good to start discussing about it, if there is a real interest of course.

### DIFF
--- a/ur_robot_driver/CMakeLists.txt
+++ b/ur_robot_driver/CMakeLists.txt
@@ -90,8 +90,6 @@ add_library(ur_robot_driver
     src/primary/robot_message.cpp
     src/primary/robot_message/version_message.cpp
     src/primary/robot_state/kinematics_info.cpp
-    src/ros/dashboard_client_ros.cpp
-    src/ros/hardware_interface.cpp
     src/rtde/control_package_pause.cpp
     src/rtde/control_package_setup_inputs.cpp
     src/rtde/control_package_setup_outputs.cpp
@@ -111,9 +109,16 @@ add_library(ur_robot_driver
 target_link_libraries(ur_robot_driver ${catkin_LIBRARIES})
 add_dependencies(ur_robot_driver ${${PROJECT_NAME}_EXPORTED_TARGETS} ${catkin_EXPORTED_TARGETS})
 
+add_library(ur_robot_driver_plugin
+  src/ros/dashboard_client_ros.cpp
+  src/ros/hardware_interface.cpp
+)
+target_link_libraries(ur_robot_driver_plugin ur_robot_driver ${catkin_LIBRARIES})
+add_dependencies(ur_robot_driver_plugin ${${PROJECT_NAME}_EXPORTED_TARGETS} ${catkin_EXPORTED_TARGETS})
+
 add_executable(ur_robot_driver_node
-  #src/ros/dashboard_client_ros.cpp
-  #src/ros/hardware_interface.cpp
+  src/ros/dashboard_client_ros.cpp
+  src/ros/hardware_interface.cpp
   src/ros/hardware_interface_node.cpp
 )
 target_link_libraries(ur_robot_driver_node ${catkin_LIBRARIES} ur_robot_driver)
--- a/ur_robot_driver/CMakeLists.txt
+++ b/ur_robot_driver/CMakeLists.txt
@@ -112,8 +112,8 @@ target_link_libraries(ur_robot_driver ${catkin_LIBRARIES})
 add_dependencies(ur_robot_driver ${${PROJECT_NAME}_EXPORTED_TARGETS} ${catkin_EXPORTED_TARGETS})
 
 add_executable(ur_robot_driver_node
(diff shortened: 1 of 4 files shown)


ur_robot_driver/CMakeLists.txt
```diff
@@ -88,6 +90,8 @@ add_library(ur_robot_driver
     src/primary/robot_message.cpp
     src/primary/robot_message/version_message.cpp
     src/primary/robot_state/kinematics_info.cpp
+    src/ros/dashboard_client_ros.cpp
+    src/ros/hardware_interface.cpp
```
fmauch: That was actually on purpose, so that the ur_robot_driver lib can be used without ROS. Since the plugin needs a library I see the problem there.

To add some perspective: I plan to move out the Non-ROS related stuff into a standalone plain cmake package. This will one day also make the ROS2 migration easier to do.

Therefore, I suggest to create another lib containing the ROS parts, namely those two source files and then link the driver_node to that as well.

ur_robot_driver/hardware_interface_plugin.xml
```diff
@@ -0,0 +1,7 @@
+<library path="lib/libur_robot_driver">
```
fmauch: we'll have to change this when we have a separate lib.

ur_robot_driver/CMakeLists.txt
```diff
@@ -108,8 +112,8 @@ target_link_libraries(ur_robot_driver ${catkin_LIBRARIES})
 add_dependencies(ur_robot_driver ${${PROJECT_NAME}_EXPORTED_TARGETS} ${catkin_EXPORTED_TARGETS})
 
 add_executable(ur_robot_driver_node
-  src/ros/dashboard_client_ros.cpp
-  src/ros/hardware_interface.cpp
+  #src/ros/dashboard_client_ros.cpp
+  #src/ros/hardware_interface.cpp
```
fmauch: See comment above.

ur_robot_driver/src/ros/hardware_interface.cpp
```diff
@@ -744,3 +744,5 @@ void HardwareInterface::publishRobotAndSafetyMode()
   }
 }
 }  // namespace ur_driver
+
+PLUGINLIB_EXPORT_CLASS(ur_driver::HardwareInterface, hardware_interface::RobotHW)
```
fmauch: We'll have to change this for the same reason as above.

gavanderhoorn:
Besides logistical issues (such as with the libraries) it would also seem the way a combined robot hw interface is orchestrated would be something to think about (as mentioned by @asierfernandez).

In the current implementation the robot controller mostly dictates control flow. I'm not sure how suitable that is when it is part of a larger composite.

ur_robot_driver/CMakeLists.txt
asierfernandez: This makes sense. I included your comments in 344e6d4cd13a27ee3cc21b2b18e81fa03d369b5e.

fmauch:
> Besides logistical issues (such as with the libraries) it would also seem the way a combined robot hw interface is orchestrated would be something to think about (as mentioned by @asierfernandez).
> 
> In the current implementation the robot controller mostly dictates control flow. I'm not sure how suitable that is when it is part of a larger composite.

Yes, we could definitely discuss this (Separately from this PR?)

We deliberately decided to let the robot dictate the timing to make real-time operation a bit easier.

@asierfernandez @gavanderhoorn Is it fine with you to merge this PR without this being tackled for the moment and we can discuss this in a separate issue?

ur_robot_driver/src/ros/hardware_interface.cpp
fmauch: I think, I was wrong there...

gavanderhoorn:
If the currently supported use-cases are still supported, we could consider merging this in as an initial effort towards `CombinedRobotHW` compatibility.

It may be nice though to explicitly document the fact that the robot controller dictates control flow for this plugin and that all `reads()` are blocking.

asierfernandez:
It's fine for me to merge it and creating a different issue #38 for discussion.
The goal of opening the issue was partially to create a discussion about this topic.

fmauch:
@asierfernandez 

#86 implements a non-blocking read. Would you agree on the implementation there?

PR #42, opened 2019-11-19
Fixup yaml-cpp linking and include paths in ur_calibration

gavanderhoorn:
`libyaml-cpp-dev` on Xenial doesn't export the `yaml::yaml` target yet, and uses a non-standard variable for it's include dirs.

These changes let Catkin properly `DEPEND` on `yaml-cpp` and correctly link to it.

### DIFF
--- a/ur_calibration/CMakeLists.txt
+++ b/ur_calibration/CMakeLists.txt
@@ -13,8 +13,6 @@ find_package(yaml-cpp REQUIRED)
 
 
 catkin_package(
-#  INCLUDE_DIRS include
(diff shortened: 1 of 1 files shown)


fmauch:
thanks for this contribution. I was always wondering how to include yaml-cpp correctly. I thought, I had it this time, but not for xenial how it seems.

gavanderhoorn:
Unless using the "modern CMake" approach it was also not entirely correct for Bionic and newer.

If/when Xenial support can be dropped, we can remove everything and just `target_link_libraries(.. yaml::yaml)`.

PR #48, opened 2019-11-27
Added a service to zero the robot's ftsensor

fmauch:
As script code interaction needs some deeper knowledge about the pitfalls of URScript, this MR provides an easy-access service to zero the robot's ftsensor.

Currently, we always advertise the service, I'd like to only advertise it when an e-Series is connected or report `false` when being called on a CB3 robot.

### DIFF
--- a/ur_robot_driver/src/ros/hardware_interface.cpp
+++ b/ur_robot_driver/src/ros/hardware_interface.cpp
@@ -317,14 +317,7 @@ bool HardwareInterface::init(ros::NodeHandle& root_nh, ros::NodeHandle& robot_hw
 
   // Calling this service will zero the robot's ftsensor. Note: On e-Series robots this will only
   // work when the robot is in remote-control mode.
-  tare_sensor_srv_ = robot_hw_nh.advertiseService<std_srvs::Trigger::Request, std_srvs::Trigger::Response>(
-      "zero_ftsensor", [&](std_srvs::Trigger::Request& req, std_srvs::Trigger::Response& resp) {
-        resp.success = this->ur_driver_->sendScript(R"(sec tareSensor():
-  zero_ftsensor()
-end
-)");
-        return true;
-      });
+  tare_sensor_srv_ = robot_hw_nh.advertiseService("zero_ftsensor", &HardwareInterface::zeroFTSensor, this);
 
   ur_driver_->startRTDECommunication();
   ROS_INFO_STREAM_NAMED("hardware_interface", "Loaded ur_robot_driver hardware_interface");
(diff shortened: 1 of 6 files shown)


ur_robot_driver/src/ros/hardware_interface.cpp
```diff
@@ -315,6 +315,19 @@ bool HardwareInterface::init(ros::NodeHandle& root_nh, ros::NodeHandle& robot_hw
   // Calling this service will make the "External Control" program node on the UR-Program return.
   deactivate_srv_ = robot_hw_nh.advertiseService("hand_back_control", &HardwareInterface::stopControl, this);
 
+  // Calling this service will zero the robot's ftsensor. Note: On e-Series robots this will only
+  // work when the robot is in remote-control mode.
+  tare_sensor_srv_ = robot_hw_nh.advertiseService<std_srvs::Trigger::Request, std_srvs::Trigger::Response>(
+      "zero_ftsensor", [&](std_srvs::Trigger::Request& req, std_srvs::Trigger::Response& resp) {
+        resp.success = this->ur_driver_->sendScript(R"(sec tareSensor():
+  set_digital_out(1,True)
```
gavanderhoorn: I'm not entirely up-to-speed on e-Series controllers: why the `set_digital_out(..)` here? Is the ft sensor connected to it?

fmauch: oops, that was from debugging...

fmauch: fixed.

ur_robot_driver/src/ros/hardware_interface.cpp
```diff
@@ -694,6 +698,28 @@ bool HardwareInterface::resendRobotProgram(std_srvs::TriggerRequest& req, std_sr
...
+{
+  if (ur_driver_->getVersion().major < 5)
+  {
+    std::stringstream ss;
+    ss << "Zeroing the Force-Torque sensor is only available for e-Series robots (Major version >= 5). This robot's "
+          "version is "
+       << ur_driver_->getVersion();
+    ROS_ERROR_STREAM(ss.str());
+    res.message = ss.str();
+    res.success = false;
```
gavanderhoorn: As `res.message` is just a `string`, would the following work?

```c++
res.message = "Zeroing the Force-Torque sensor is only available for e-Series robots (Major version >= 5). This robot's version is " + ur_driver_->getVersion();
ROS_ERROR_STREAM(res.message);
```

Avoids creating a `std::stringstream`.

fmauch: Not quite, as `UrDriver::getVersion()` returns an object of type `VersionInformation` which has a `<<` operator (also implemented in this PR). With your suggestion we would have to add a `toString()` method or similar to `VersionInformation` instead. Personally, I like using stream operators for this. Is there any particular reason why you would prefer avoiding the `std::stringstream`?

gavanderhoorn: It's a pattern I've used before. Hence the suggestion.

I'm not against using `std::stringstream` here.

PR #49, opened 2019-11-27
Always end script commands with a newline

fmauch:
Otherwise script will not be interpreted by the robot which might be counter-intuitive.
Changing the behavior as such will also be the same as in the ur_modern_driver
so migrating will be easier.

I decided to change the function's interface to copy the string in order add a
trailing '\n' if necessary.

### DIFF
--- a/ur_robot_driver/src/ur/ur_driver.cpp
+++ b/ur_robot_driver/src/ur/ur_driver.cpp
@@ -266,6 +266,9 @@ bool UrDriver::sendScript(const std::string& program)
                              "should not happen.");
   }
 
+  // urscripts (snippets) must end with a newline, or otherwise the controller's runtime will
+  // not execute them. To avoid problems, we always just append a newline here, even if
+  // there may already be one.
   auto program_with_newline = program + '\n';
 
   size_t len = program_with_newline.size();
--- a/ur_robot_driver/src/ur/ur_driver.cpp
+++ b/ur_robot_driver/src/ur/ur_driver.cpp
@@ -266,13 +266,15 @@ bool UrDriver::sendScript(const std::string& program)
                              "should not happen.");
   }
 
-  size_t len = program.size();
-  const uint8_t* data = reinterpret_cast<const uint8_t*>(program.c_str());
+  auto program_with_newline = program + '\n';
+
+  size_t len = program_with_newline.size();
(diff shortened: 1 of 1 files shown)


ur_robot_driver/src/ur/ur_driver.cpp
```diff
@@ -258,21 +258,25 @@ rtde_interface::RTDEWriter& UrDriver::getRTDEWriter()
...
                              "should not happen.");
   }
 
-  size_t len = program.size();
   const uint8_t* data = reinterpret_cast<const uint8_t*>(program.c_str());
+  if (program.back() != '\n')
+  {
+    program.append("\n");
+  }
+  size_t len = program.size();
```
gavanderhoorn: Are scripts with empty lines not legal?

If they are, I would suggest something like this:

```c++
auto program_with_newline = program + '\n';
```
and then do whatever was done before.

fmauch: looks cleaner, I agree. Scripts with multiple empty lines should be working, as well. (Just tested to make sure)

ur_robot_driver/src/ur/ur_driver.cpp
```diff
@@ -266,13 +266,15 @@ bool UrDriver::sendScript(const std::string& program)
                              "should not happen.");
   }
 
+  auto program_with_newline = program + '\n';
+
   size_t len = program.size();
   const uint8_t* data = reinterpret_cast<const uint8_t*>(program.c_str());
```
gavanderhoorn: Wouldn't you need to replace `program` here with `program_with_newline` as well?

fmauch: indeed...

ur_robot_driver/src/ur/ur_driver.cpp
```diff
@@ -266,13 +266,15 @@ bool UrDriver::sendScript(const std::string& program)
                              "should not happen.");
   }
 
-  size_t len = program.size();
-  const uint8_t* data = reinterpret_cast<const uint8_t*>(program.c_str());
+  auto program_with_newline = program + '\n';
```
gavanderhoorn:
```suggestion
  // urscripts (snippets) must end with a newline, or otherwise the controller's runtime will
  // not execute them. To avoid problems, we always just append a newline here, even if
  // there may already be one.
  auto program_with_newline = program + '\n';
```

PR #50, opened 2019-11-27
Added a service to setup the active payload

fmauch:
This PR adds a service to setup the active payload through a ROS service. 

This PR depends on https://github.com/ros-industrial/universal_robot/pull/470 being accepted which is why this is a draft PR.

### DIFF
--- a/ur_robot_driver/src/ros/hardware_interface.cpp
+++ b/ur_robot_driver/src/ros/hardware_interface.cpp
@@ -389,6 +389,7 @@ bool HardwareInterface::init(ros::NodeHandle& root_nh, ros::NodeHandle& robot_hw
   set_payload_srv_ = robot_hw_nh.advertiseService<ur_msgs::SetPayload::Request, ur_msgs::SetPayload::Response>(
       "set_payload", [&](ur_msgs::SetPayload::Request& req, ur_msgs::SetPayload::Response& resp) {
         std::stringstream cmd;
+        cmd.imbue(std::locale::classic());  // Make sure, decimal divider is actually '.'
         cmd << "sec setup():" << std::endl
             << " set_payload(" << req.payload << ", [" << req.center_of_gravity.x << ", " << req.center_of_gravity.y
             << ", " << req.center_of_gravity.z << "])" << std::endl
--- a/ur_robot_driver/src/ros/hardware_interface.cpp
+++ b/ur_robot_driver/src/ros/hardware_interface.cpp
@@ -29,6 +29,8 @@
 #include "ur_robot_driver/ur/tool_communication.h"
 #include <ur_robot_driver/exceptions.h>
 
+#include <ur_msgs/SetPayload.h>
+
 #include <Eigen/Geometry>
 
 using industrial_robot_status_interface::RobotMode;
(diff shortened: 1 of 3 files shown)


ur_robot_driver/src/ros/hardware_interface.cpp
```diff
@@ -318,6 +320,18 @@ bool HardwareInterface::init(ros::NodeHandle& root_nh, ros::NodeHandle& robot_hw
   ur_driver_->startRTDECommunication();
   ROS_INFO_STREAM_NAMED("hardware_interface", "Loaded ur_robot_driver hardware_interface");
 
+  // Setup the mounted payload through a ROS service
+  set_payload_srv_ = robot_hw_nh.advertiseService<ur_msgs::SetPayload::Request, ur_msgs::SetPayload::Response>(
+      "set_payload", [&](ur_msgs::SetPayload::Request& req, ur_msgs::SetPayload::Response& resp) {
+        std::stringstream cmd;
+        cmd << "sec setup():" << std::endl
+            << " set_payload(" << req.payload << ", [" << req.center_of_gravity.x << ", " << req.center_of_gravity.y
+            << ", " << req.center_of_gravity.z << "])" << std::endl
```
gavanderhoorn: Is relying on `<<` to convert to `string` locale safe?

Could end up with multiple commas in the string, which would probably upset UR's parser.

fmauch: I did not think about that. I would never have though about the stream operator using locales in the background.

In a quick example I could not get a stream into locale-like formatting without explicitly setting the locale of the stream. However, to make sure, we could set the string's locale explicitly to [`classic`](https://en.cppreference.com/w/cpp/locale/locale/classic)

gavanderhoorn: I just want to avoid another ros/urdfdom_headers#41.

If stream io doesn't suffer from this, all the better.

ethanfowler:
Is there anything holding this up that I can help with? We (Shadow robot) want this functionality as soon as possible, though we can run on a branch in the meantime.

lavarock1234:
Any update on this PR?  Thanks very much in advance

gavanderhoorn:
The referenced PR has been merged into [ros-industrial/ur_msgs](https://github.com/ros-industrial/ur_msgs).

@fmauch: I'd appreciate it if you could test the service a bit before I release it into Melodic.

fmauch:
> The referenced PR has been merged into [ros-industrial/ur_msgs](https://github.com/ros-industrial/ur_msgs).
> 
> @fmauch: I'd appreciate it if you could test the service a bit before I release it into Melodic.

Sure thing!

fmauch:
Just tested with a 
```
rosservice call /ur_hardware_interface/set_payload "payload: 2.0
center_of_gravity:
  x: 0.2
  y: 0.3
  z: 0.1"
```

and verified with

```
rostopic pub /ur_hardware_interface/script_command std_msgs/String "data: 'def myProgram():

 textmsg(get_target_payload())

 textmsg(get_target_payload_cog())

end'"
```

Which printed
image

fmauch:
It would be nice if we could add a check for success into the service. Currently, it succeeds, as soon as the script code got successfully sent to the socket.

Unfortunately, this isn't very easy to do as far as I understand. To my knowledge, there is no interface to get information about the payload data in
 - the primary interface
 - the RTDE interface
 - the dashboard interface

It could be implemented together with #171, where we would receive a `PROGRAM_XXX_STARTED`, `PROGRAM_XXX_STOPPED` message sequence on success  and a `PROGRAM_XXX_STARTED`, `RuntimeError`, `PROGRAM_XXX_STOPPED` sequence in case of illegal input. Timing out on an answer should also be a failure.

Another option would be of course to send payload data back through a custom socket, but then we would be highly dependent on our actual program script code.

gavanderhoorn:
Update: `ur_msgs` was released into Melodic (`v1.3.1`) and includes the `SetPayload` service definition.

PR #60, opened 2019-12-11
robot_driver: update tracker and repo urls.

gavanderhoorn:
As per subject.

Noticed it when I was looking at the `ur_robot_driver` wiki page.


fmauch:
That went wrong during search&replace, thank you!

PR #63, opened 2019-12-12
Fix spelling of "actual_main_voltage"

cjue:
Only concerned initialisation. Name not used elsewhere in ur_robot_driver.


fmauch:
Thanks!

PR #35, opened 2019-11-12
Rtde recipe publishing

t-schnell:
Implements automatic publishing of requested RTDE data to ROS topics.

PR #72, opened 2019-12-21
Added a topic that shows program data for any program

dzoberti:
This can be useful for a number of reasons. Mainly it allows to detect when the robot has finished executing a script received on the `script_command` topic


fmauch:
Thank you very much for your contribution. However, please be aware that this is already part of #35 if I understand this correctly. If that is the case, we will probably not merge this, but it can serve other users as a good base to achieve the same until #35 is finished.

PR #75, opened 2020-01-09
fixed duplicated service advertisements

mgkwon:
This PR fixes the duplicated assignments to a same ROS service server object.


fmauch:
@mgkwon Thank you for your contribution!

After checking each and every service inside the dashboard client again, I noticed that the `polyscope_version_service_` is not implemented either. Would you mind me pushing this into this PR, as well?

fmauch:
There are a couple of calls not implemented (even more since 3.12/5.6) I'll merge this as is and add missing calls at a later stage.

PR #86, opened 2020-01-23
Adding non blocking read

carebare47:
Adds functionality to enable non blocking read. 

When this new driver is used with other robots via combined_robot_hw, the 100ms timeout on the blocking read operation slows the loop rate of the other robots. This PR adds a parameter to set this timeout to zero, making the read non-blocking. It also disables the error generated when read returns without data. Finally, it synchronises read/write when this parameter is enabled.

### DIFF
--- a/ur_robot_driver/doc/ROS_INTERFACE.md
+++ b/ur_robot_driver/doc/ROS_INTERFACE.md
@@ -32,6 +32,10 @@ Standalone launchfile to startup a ur3e. This requires a robot reachable via a n
 
     Use the description in limited mode (Every axis rotates from -PI to PI)
 
+ * "**reverse_port**" (default: "50001")
+
+    Port that will be opened by the driver to allow direct communication between the driver and the robot controller.
+
  * "**robot_description_file**" (default: "$(find ur_e_description)/launch/ur3e_upload.launch")
 
     Robot description launch file.
@@ -39,6 +43,10 @@ Standalone launchfile to startup a ur3e. This requires a robot reachable via a n
  * "**robot_ip**" (Required)
 
     IP address by which the robot can be reached.
+
+ * "**script_sender_port**" (default: "50002")
+
+    The driver will offer an interface to receive the program's URScript on this port. If the robot cannot connect to this port, `External Control` will stop immediately.
 
  * "**stopped_controllers**" (default: "pos_traj_controller")
 
@@ -113,6 +121,10 @@ Standalone launchfile to startup a ur10 robot. This requires a robot reachable v
 
     Use the description in limited mode (Every axis rotates from -PI to PI)
 
+ * "**reverse_port**" (default: "50001")
+
+    Port that will be opened by the driver to allow direct communication between the driver and the robot controller.
+
  * "**robot_description_file**" (default: "$(find ur_description)/launch/ur10_upload.launch")
 
     Robot description launch file.
@@ -120,6 +132,10 @@ Standalone launchfile to startup a ur10 robot. This requires a robot reachable v
  * "**robot_ip**" (Required)
 
     IP address by which the robot can be reached.
+
+ * "**script_sender_port**" (default: "50002")
+
+    The driver will offer an interface to receive the program's URScript on this port. If the robot cannot connect to this port, `External Control` will stop immediately.
 
  * "**stopped_controllers**" (default: "pos_traj_controller")
 
@@ -158,6 +174,10 @@ Robot bringup launchfile without the robot description. Include this, if you wan
 
     Please add description. See file "launch/ur_control.launch".
 
+ * "**reverse_port**" (default: "50001")
+
+    Port that will be opened by the driver to allow direct communication between the driver and the robot controller.
+
  * "**robot_ip**" (Required)
 
     IP address by which the robot can be reached.
@@ -169,6 +189,10 @@ Robot bringup launchfile without the robot description. Include this, if you wan
  * "**rtde_output_recipe_file**" (default: "$(find ur_robot_driver)/resources/rtde_output_recipe.txt")
 
     Recipe file used for the RTDE-outputs. Only change this if you know what you're doing.
+
+ * "**script_sender_port**" (default: "50002")
+
+    The driver will offer an interface to receive the program's URScript on this port. If the robot cannot connect to this port, `External Control` will stop immediately.
 
  * "**stopped_controllers**" (default: "pos_traj_controller")
 
@@ -247,6 +271,10 @@ Launchfile that starts a robot description with robot_state publisher and the dr
 
     Use the description in limited mode (Every axis rotates from -PI to PI)
 
+ * "**reverse_port**" (default: "50001")
+
+    Port that will be opened by the driver to allow direct communication between the driver and the robot controller.
+
  * "**robot_description_file**" (Required)
 
     Robot description launch file.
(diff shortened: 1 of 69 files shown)


ur_robot_driver/src/ros/hardware_interface.cpp
```diff
@@ -104,6 +104,9 @@ bool HardwareInterface::init(ros::NodeHandle& root_nh, ros::NodeHandle& robot_hw
     return false;
   }
 
+  // Enables non_blocking_read mode. Useful when used with combined_robot_hw
```
fmauch: We could improve documentation a bit, here. I would especially note that error output on missing packets are suppressed.

ur_robot_driver/src/ros/hardware_interface.cpp
```diff
@@ -431,15 +435,18 @@ void HardwareInterface::read(const ros::Time& time, const ros::Duration& period)
...
+    }
   }
 }
 
 void HardwareInterface::write(const ros::Time& time, const ros::Duration& period)
 {
   if ((runtime_state_ == static_cast<uint32_t>(rtde_interface::RUNTIME_STATE::PLAYING) ||
        runtime_state_ == static_cast<uint32_t>(rtde_interface::RUNTIME_STATE::PAUSING)) &&
-      robot_program_running_)
+      robot_program_running_ && ((non_blocking_read_ && packet_read_) || (!non_blocking_read_)))
```
fmauch: Could this be reduced to 
```suggestion
      robot_program_running_ && packet_read_
```

If we are using blocking read then we still set `packet_read_` to `true` in each cycle, right?

ur_robot_driver/src/ur/ur_driver.cpp
```diff
@@ -141,8 +144,11 @@ ur_driver::UrDriver::UrDriver(const std::string& robot_ip, const std::string& sc
 
 std::unique_ptr<rtde_interface::DataPackage> ur_driver::UrDriver::getDataPackage()
 {
-  std::chrono::milliseconds timeout(100);  // We deliberately have a quite large timeout here, as the robot itself
-                                           // should command the control loop's timing.
+  std::chrono::milliseconds timeout(get_packet_timeout_);  // This can take one of two values, 0ms or 100ms. The large
+                                                           // timeout is for when the robot is commanding the control
+                                                           // loop's timing (read is blocking). The zero timeout is for
+                                                           // when the robot is sharing a control loop with
+                                                           // something else (combined_robot_hw)
```
fmauch: We could move the comment above the line to improve readability.

ur_robot_driver/src/ros/hardware_interface.cpp
carebare47: How does that look now?

ur_robot_driver/src/ros/hardware_interface.cpp
carebare47: If I've understood correctly, that would force read and write to be synchronised whether or not the non_blocking_read parameter was set. At the moment there's nothing that forces one write for every one read. When used with combined_robot_hw we get eight writes (probably containing the same data) for every one read due to the arm running at 125Hz and the other robot in the loop running at 1KHz. It could be simplified to: 
```suggestion
      robot_program_running_ && (!non_blocking_read_ || packet_read_))
```
but I thought the current logic was a little more readable.

ur_robot_driver/src/ur/ur_driver.cpp
carebare47: How's this? https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/86/commits/529c711c22bb58167a7c5b195dbffc3dd038b2c9

ur_robot_driver/src/ros/hardware_interface.cpp
carebare47: How about this? https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/commit/529c711c22bb58167a7c5b195dbffc3dd038b2c9

ur_robot_driver/src/ur/ur_driver.cpp
carebare47: Like that? https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/86/commits/4cf8d6fc8d8c8240008c319659777e5f55d546b6

carebare47: Like that? https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/86/commits/4cf8d6fc8d8c8240008c319659777e5f55d546b6

ur_robot_driver/src/ros/hardware_interface.cpp
carebare47: Although looking at them next to each other, I do prefer ` robot_program_running_ && (!non_blocking_read_ || packet_read_))
`

fmauch: I also like your second suggestion the most.

carebare47: https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/86/commits/3a8eea838eec04d5fae6405b4135f251a2a1a6ba

fmauch:
@carebare47  I added a PR with generated documentation for the new parameter on your branch.

asierfernandez:
This implementation is also fine for us, clean and efficient. 
I've tested under the offline simulator with combined_robot_hw and it works as I'd expect.

carebare47:
> @carebare47 I added a PR with generated documentation for the new parameter on your branch.

Thanks for doing that, I've merged it into my branch :)

ur_robot_driver/doc/ROS_INTERFACE.md
```diff
@@ -762,6 +762,10 @@ This is the actual driver node containing the ROS-Control stack. Interfaces docu
 
     Hash of the calibration reported by the robot. This is used for validating the robot description is using the correct calibration. If the robot's calibration doesn't match this hash, an error will be printed. You can use the robot as usual, however Cartesian poses of the endeffector might be inaccurate. See the "ur_calibration" package on help how to generate your own hash matching your actual robot.
 
+ * "**non_blocking_read**" (default: "false")
+
+    Enables non_blocking_read mode. Useful when used with combined_robot_hw. Disables error generated when read returns without any data, sets the read timeout to zero, and synchronises read/write operations.
+
```
gavanderhoorn: What about using this without `combined_robot_hw`? Should that be discouraged / made impossible?

carebare47: How about "Enables non_blocking_read mode. Should only be used with combined_robot_hw. Disables error generated when read returns without any data, sets the read timeout to zero, and synchronises read/write operations. Enabling this when not used with combined_robot_hw can suppress important errors and affect real-time performance." ?

fmauch: I would be fine with that.

carebare47: https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/86/commits/1f011214b29d3a31f1ca0193a0a1cbe5b5599988

fmauch:
thanks again @carebare47

PR #89, opened 2020-01-25
config: use yaml anchor to reduce magic nrs

gavanderhoorn:
As per subject.

Almost like a named constant, but in yaml.


fmauch:
I like! Thanks @gavanderhoorn

PR #93, opened 2020-01-28
Close all closable sockets

### LINKED ISSUE #91: Filehandles of TCP connection not closed when "connection to robot dropped"
# Summary

Everytime the externalControl is activated the ur_robot_driver opens a new socket (and seems to not close it when the connection is dropped).
This eventually leads to a "too many files open" exception and the crash of the driver.
The ur program continues to run and seems to skip over (??) the external_control urcap (to be verified).

# Versions
 - ROS Driver version: [master](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/commit/f2db8f3d000c24b77e76a2a2ae89ac2feaf7dc1c)
 - Tested with Robot Software Version(s): ursim 3.7

# Impact
With this issue this can not be used in an industrial (production line) setup. It will eventually stop and the driver needs to restart.

## Steps to Reproduce
* start the ros_controller (ur_robot_driver)
* start another node, that waits for "true" on rostopic "/ur_hardware_interface/robot_program_running"; if so sleep for 0.5s; then call rosservice "/ur_hardware_interface/hand_back_control"

* start a ur program with an infinite loop with "ExternalControl"

## Expected Behaviour
Resources of the program should be stable and not grow over time.

## Actual Behavior
watch -n 0.5 cat /proc/_{PID of ur_robot_driver}_/net/sockstat

sockets: used 3082 **[continuously increasing]**
TCP: inuse 1056  **[continuously increasing]** orphan 0 tw 295 alloc 1057 mem 211
UDP: inuse 18 mem 27
UDPLITE: inuse 0
RAW: inuse 0
FRAG: inuse 0 memory 0

fmauch:
Sockets do not necessarily have to be in state connected when they should be
closed. Before, only connected sockets got closed leading to a "socket leak"
if a socket was disconnected before a close request was processed.

With this fix all sockets with a valid file descriptor get closed when close()
is being called.

This fixes #91

### DIFF
--- a/ur_robot_driver/src/comm/server.cpp
+++ b/ur_robot_driver/src/comm/server.cpp
@@ -102,9 +102,6 @@ bool URServer::accept()
 
 void URServer::disconnectClient()
 {
-  if (client_.getState() != comm::SocketState::Connected)
-    return;
-
   client_.close();
 }
 
--- a/ur_robot_driver/src/comm/tcp_socket.cpp
+++ b/ur_robot_driver/src/comm/tcp_socket.cpp
@@ -118,11 +118,12 @@ bool TCPSocket::setSocketFD(int socket_fd)
 
 void TCPSocket::close()
 {
-  if (state_ != SocketState::Connected)
-    return;
-  state_ = SocketState::Closed;
-  ::close(socket_fd_);
-  socket_fd_ = -1;
+  if (socket_fd_ >= 0)
+  {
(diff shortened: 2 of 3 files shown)


ipa-lth:
This PR resolves the issue. I rerun the continues test for about 4 hours without any problems or critical tendencies.

fmauch:
Then I'll got ahead and merge this.

PR #104, opened 2020-02-11
Require a controller reset when reading data from RTDE fails

fmauch:
Otherwise the joint_state_controller will continue publishing old joint data.

### DIFF
--- a/ur_robot_driver/src/ros/hardware_interface.cpp
+++ b/ur_robot_driver/src/ros/hardware_interface.cpp
@@ -466,6 +466,7 @@ void HardwareInterface::read(const ros::Time& time, const ros::Duration& period)
   {
     if (!non_blocking_read_)
     {
+      controller_reset_necessary_ = true;
       ROS_ERROR("Could not get fresh data package from robot");
     }
   }
--- a/ur_robot_driver/src/ros/hardware_interface.cpp
+++ b/ur_robot_driver/src/ros/hardware_interface.cpp
@@ -464,6 +464,13 @@ void HardwareInterface::read(const ros::Time& time, const ros::Duration& period)
   }
   else
   {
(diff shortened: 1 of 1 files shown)


ur_robot_driver/src/ros/hardware_interface.cpp
```diff
@@ -464,6 +464,7 @@ void HardwareInterface::read(const ros::Time& time, const ros::Duration& period)
   }
   else
   {
+    controller_reset_necessary_ = true;
```
gavanderhoorn: Would perhaps be nice to add a comment explaining why we're resetting them here in this case?

fmauch: True... Added in ca63af0

gavanderhoorn: Just wondering: would there be any way we could notify the user of this happening? Any `printf(..)` (or similar) would break determinism, but I'm wondering whether it would be worth it.

fmauch: Well, in the blocking (default) case we have the error output already. The more I think about it: In the blocking case we actually do not really want this behavior right? (assuming that anybody using this driver inside a combined HW calls the `HardwareInterface::shouldResetControllers()` method.

gavanderhoorn: In the blocking case I assume we just don't get past the `read()`?

re: non-blocking/combined-hw: yeah, this is one of the aspects which I don't really like about combined hw. It just doesn't integrate as nicely with everything, as the hw-iface has to give up some control.

fmauch: The blocking case also has a time out which will trigger the error that no new data could be read.

PR #109, opened 2020-02-13
F ignore empty list in do switch

carebare47:
Currently, if doSwitch is called with empty lists for start_list and stop_list, it will stop communication to an arm. This is a problem when used with combined_robot_hw, as combined_robot_hw will call the doSwitch method of all instances of hardware_interface when a controller is started, but only pass one instance a list containing a controller, stopping any other instances started by combined_robot_hw. This pr makes sure that either the start_list or the stop_list contains something before halting communications.


fmauch:
Thanks for pointing that out. Actually, I think this problem even goes beyond that. Matching controllers only by the type of their claimed resources isn't very robust.

If it's ok with you, I would rather use this opportunity to get this right.

I think, we should
 * not set any running state to false initially, but only act on start and stop lists
 * match controllers to the actual hardware interface

See #111 for my alternative implementation.

carebare47:
Thanks for your help with this. Your approach looks much more robust, happy to close this :)

PR #123, opened 2020-02-26
Add missing package dependency

mintar:


gavanderhoorn:
This is indeed used by `ur_robot_driver`, but seeing as it's hosted in the same repository, I'm curious as to when not having this declaration there could lead to problems.

@mintar: could you provide some more information on what made you submit this PR?

fmauch:
It seems to be right to me. Isn't it common practice to also include those from the same repository. As soon as we will have a binary release this will matter for sure, right?

mintar:
> @mintar: could you provide some more information on what made you submit this PR?

I like to use [roslaunch_add_file_check](http://wiki.ros.org/roslaunch#Catkin) in my CMakeLists. When I include one of the launch files from ur_robot_driver in mine, it leads to a test failure.

Here's a minimal working example:

**temp.launch**
```xml
<launch>
  <include file="$(find ur_robot_driver)/launch/ur5_bringup.launch">
    <arg name ="robot_ip" value="foo" />
  </include>
</launch>
```

```bash
$ rosrun roslaunch roslaunch-check temp.launch                                                                  
checking temp.launch
Missing package dependencies: ur_robot_driver/package.xml: controller_stopper
...writing test results to /home/martin/.ros/test_results/ur_robot_driver/rosunit-roslaunch_check_launch_temp_launch.xml
FAILURE:
[temp.launch]:
	Missing package dependencies: ur_robot_driver/package.xml: controller_stopper
wrote test file to [/home/martin/.ros/test_results/ur_robot_driver/rosunit-roslaunch_check_launch_temp_launch.xml]
```

I agree that since controller_stopper is hosted in the same repo, the missing dependency usually shouldn't cause any problems. Unless you plan to release it later: then you need the package.xml dependency in order to generate a debian package dependency.

gavanderhoorn:
As I wrote earlier: the suggested change is the right one, as the dependency declaration is indeed missing.

I just wanted to understand better in which context you ran into an error or problem caused by this.

Thanks for clarifying.

PS: `catkin_lint` also checks for these kinds of things with the recent release. It may be something to add to the CI configuration @fmauch.

PR #132, opened 2020-03-05
Fix package dependencies

fmauch:
Builds ontop of #123 

This should
 - Use `catkin_lint` in the ci pipeline
 - Fix all package dependencies

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,8 @@ env:
   global:
     - CCACHE_DIR=$HOME/.ccache
   matrix:
-    - ROS_DISTRO="kinetic" ROS_REPO=ros UPSTREAM_WORKSPACE=file CATKIN_LINT=true
-    - ROS_DISTRO="melodic" ROS_REPO=ros UPSTREAM_WORKSPACE=file CATKIN_LINT=true
+    - ROS_DISTRO="kinetic" ROS_REPO=ros UPSTREAM_WORKSPACE=file
+    - ROS_DISTRO="melodic" ROS_REPO=ros UPSTREAM_WORKSPACE=file
     - ROS_DISTRO="kinetic" CLANG_FORMAT_CHECK=file CLANG_FORMAT_VERSION=6.0
 
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,8 @@ env:
   global:
     - CCACHE_DIR=$HOME/.ccache
   matrix:
-    - ROS_DISTRO="kinetic" ROS_REPO=ros UPSTREAM_WORKSPACE=file
-    - ROS_DISTRO="melodic" ROS_REPO=ros UPSTREAM_WORKSPACE=file
+    - ROS_DISTRO="kinetic" ROS_REPO=ros UPSTREAM_WORKSPACE=file CATKIN_LINT=true
+    - ROS_DISTRO="melodic" ROS_REPO=ros UPSTREAM_WORKSPACE=file CATKIN_LINT=true
     - ROS_DISTRO="kinetic" CLANG_FORMAT_CHECK=file CLANG_FORMAT_VERSION=6.0
 
 install:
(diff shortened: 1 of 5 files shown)


fmauch:
`catkin_lint` fails on yaml-cpp, where I don't see why. In #42 defining the include dir variable should have fixed that, but `catkin_lint` says:

```
ur_calibration: CMakeLists.txt(17): error: catkin_package() lists unconfigured system package 'YAML_CPP'
     * In order to export a system package as dependency, you must
     * either call find_package(YAML_CPP) first or initialize the
     * YAML_CPP_INCLUDE_DIRS and YAML_CPP_LIBRARIES variables
     * manually.
     * You can ignore this problem with --ignore unconfigured_system_depend
catkin_lint: checked 5 packages and found 1 problems
```

However, the variables are initialized manually:

```cmake
# create alias for the '_INCLUDE_DIR' variable, to help Catkin export it
set(YAML_CPP_INCLUDE_DIRS ${YAML_CPP_INCLUDE_DIR})

message(WARNING "YAML_CPP_INCLUDE_DIRS: ${YAML_CPP_INCLUDE_DIRS}")
message(WARNING "YAML_CPP_LIBRARIES: ${YAML_CPP_LIBRARIES}")


catkin_package(
  CATKIN_DEPENDS
    roscpp
    ur_robot_driver
  DEPENDS
    YAML_CPP
)
```
leads to
```
CMake Warning at ur_rtde_driver/ur_calibration/CMakeLists.txt:17 (message):
  YAML_CPP_INCLUDE_DIRS:
  /usr/lib/x86_64-linux-gnu/cmake/yaml-cpp/../../../../include


CMake Warning at ur_rtde_driver/ur_calibration/CMakeLists.txt:18 (message):
  YAML_CPP_LIBRARIES: yaml-cpp
```

@gavanderhoorn Do you know how to deal with this?

gavanderhoorn:
So this appears to be a bit of an either/or.

I've submitted https://github.com/fkie/catkin_lint/issues/78 for some discussion with @roehling.

gavanderhoorn:
So `yaml-cpp` is now treated special by `catkin_lint`. It should no longer flag `YAML_CPP` as a problem if `find_package(yaml-cpp ..)` is used earlier.

I'm not entirely sure where `industrial_ci` get `catkin_lint` from, so it may take some time for this job to start turning green.

roehling:
I had a look at [industrial_ci](https://github.com/ros-industrial/industrial_ci) and I believe they install `catkin_lint` from `packages.ros.org`. Unfortunately, the OSRF people stopped updating at version 1.5.4 (November 2018), so that version is rather outdated by now.  The [moveit_ci](https://github.com/ros-planning/moveit_ci) people switched to PyPI, which is generally the first place where new versions show up.

fmauch:
> I had a look at [industrial_ci](https://github.com/ros-industrial/industrial_ci) and I believe they install `catkin_lint` from `packages.ros.org`. Unfortunately, the OSRF people stopped updating at version 1.5.4 (November 2018), so that version is rather outdated by now. The [moveit_ci](https://github.com/ros-planning/moveit_ci) people switched to PyPI, which is generally the first place where new versions show up.

This was explicitly changed in https://github.com/ros-industrial/industrial_ci/pull/448 (probably, as python2 support ran out) @roehling Do you know whether there is any particular reason why there is no more recent released (as inside the OSRF deb repos) version?

roehling:
> Do you know whether there is any particular reason why there is no more recent released (as inside the OSRF deb repos) version?

As far as I can tell, it is simply a matter of priorities and available maintenance time. Most if not all OSRF guys seem to be working on ROS2, and for a new `catkin_lint` release, someone has to trigger an import manually, AFAIK. I used to open issues for that, but I stopped doing so after a while.

Given that `catkin_lint` has become part of Debian upstream and the latest release is always available from PyPI and my PPA, I felt that there's no need to bother OSRF any more.

gavanderhoorn:
Thing is: ROS users are actively discouraged from using `pip` to install "ROS packages" or their dependencies, and adding your PPA is an extra step.

All ROS users have the ROS repositories configured and active.

roehling:
I see your point and [opened an issue](https://github.com/ros/rosdistro/issues/24251) to import the latest release (which has the yaml-cpp fix).

gavanderhoorn:
I've added an additional comment to your issue over there.

fmauch:
I reverted the catkin_lint commit to add this again at a later stage.

PR #139, opened 2020-03-21
real_time.md improvements

sharpe-developer:
Adding these items to the real-time patch installation instructions eliminate issues encountered when installing on a clean Ubuntu 18.04 system.

### DIFF
--- a/ur_robot_driver/doc/real_time.md
+++ b/ur_robot_driver/doc/real_time.md
@@ -17,7 +17,7 @@ To build the kernel, you will need a couple of tools available on your system. Y
 using
 
 ``` bash
-$ sudo apt-get install build-essential bc ca-certificates gnupg2 libssl-dev wget gawk
+$ sudo apt-get install build-essential bc ca-certificates gnupg2 libssl-dev wget gawk flex
 ```
 
 Before you download the sources of a real-time-enabled kernel, check the kernel version that is currently installed:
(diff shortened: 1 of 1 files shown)


fmauch:
Could you please elaborate your reasoning a bit further?

sharpe-developer:
> Could you please elaborate your reasoning a bit further?

Absolutely! I was following the instructions in `real_time.md` to install the real-time patch on a clean install of Ubuntu 18.04.03. While following the instructions as written I encountered the following issues which caused errors in the procedure. 

1.  When unzipping the patch file `xz -d patch-4.14.139-rt66.patch.xz` the `xz -d` command extracts the file but removes the original compressed file `patch-4.14.139-rt66.patch.xz` file. In a later step the patch is applied using the `xz` file `xzcat ../patch-4.14.139-rt66.patch.xz | patch -p1`. As you can see this command expects the `patch-4.14.139-rt66.patch.xz` file to be present in the directory. However, the file is not present because of the earlier `xz -d` command. Adding the `-k` option to the `xz` command extracts the file but also leaves the original compressed file in place.

2. When going through the process the process failed (during `make oldconfig` I think) because flex and bison were not installed. Installing these packages during the `apt-get install` step allows  `make oldconfig` to execute without failing due to missing packages.

The changes I have proposed allow the process in to `real_time.md` to be executed without any errors on an Ubuntu 18.04 system after clean install of Ubuntu.

fmauch:
Thank you!

gavanderhoorn:
If `patch` is getting the compressed version of patch directly (via `xzcat`), why decompress it?

sharpe-developer:
The instructions decompress the file for use when verifying the signature of the patch.

Issue #32, opened 2019-11-07
Publish industrial_msgs/RobotStatus messages


gavanderhoorn:
tl;dr: would there be interest in adding [industrial_robot_status_controller](http://wiki.ros.org/industrial_robot_status_controller) compatibility to `ur_robot_driver`?

(I've ignored the template, as it doesn't seem to cater to discussions about potential contributions / feature/enhancement requests)

---

`ur_modern_driver` publishes [industrial_msgs/RobotStatus](http://docs.ros.org/api/industrial_msgs/html/msg/RobotStatus.html) messages which give a (admittedly limited) view on robot/controller state.

`ur_robot_driver` does not do this, which complicates migrating from `ur_modern_driver`.

Instead of adding support for this directly to `ur_robot_driver`, an intermediary in the form of the [industrial_robot_status_controller](http://wiki.ros.org/industrial_robot_status_controller) could be used. `ur_robot_driver` would have to expose an `IndustrialRobotStatusHandle` and populate some fields in that, after which the controller would be responsible for conversion to and publication of the actual `RobotStatus` message.

As without the controller nothing happens, users would have an opt-in and the default behaviour of the `hardware_interface` would not change.

fmauch:
Yes, this seems very reasonable to me :-)

gavanderhoorn:
I'll see if I can submit a PR for this in the next few days.

axelschroth:
Any plans/progress on this?

gavanderhoorn:
It's on my TODO list, but I've not worked on it yet.

axelschroth:
Since we are using robot status messages and need this to be implemented, do you have a rough estimate when you'll be able to submit a PR? Can I support you in any way?

gavanderhoorn:
If you need this, then please don't wait on me, as I will not have time to work on this in the foreseeable future.

IIRC, @simonschmeisser has already implemented this once in `ur_modern_driver`. The idea is the same, but now you'd use the intermediate `IndustrialRobotStatusInterface` to push all the info into. The `IndustrialRobotStatusController` will do the rest.

The readme of `industrial_robot_status_controller` should contain enough info on this ([here](https://github.com/gavanderhoorn/industrial_robot_status_controller#usageintegration)).

axelschroth:
Alright, thank you! I'll have a look into this issue shortly.

axelschroth:
[...]

PR #141, opened 2020-03-24
Robot status topic via controller

axelschroth:
Implements issue https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/issues/32

Notes and limitations:
* controller publishes state messages as long as ExternalControl is running or paused on UR. Otherwise `rostopic echo /robot_status` does not receive further messages.
* Tested with UR simulation software -> no tests regarding emergency stops (since this is not possible with ursim)
* Criteria of published content is based on ur_modern_driver implementation (see https://github.com/Zagitta/ur_modern_driver/pull/5)

### DIFF
--- a/ur_robot_driver/config/ur3e_controllers.yaml
+++ b/ur_robot_driver/config/ur3e_controllers.yaml
@@ -1,6 +1,6 @@
 # Settings for ros_control control loop
 hardware_control_loop:
-   loop_hz: 500
+   loop_hz: &loop_hz 500
 
 # Settings for ros_control hardware interface
 hardware_interface:
@@ -15,17 +15,17 @@ hardware_interface:
 # Publish all joint states ----------------------------------
 joint_state_controller:
    type:         joint_state_controller/JointStateController
-   publish_rate: 500
+   publish_rate: *loop_hz
 
 # Publish wrench ----------------------------------
 force_torque_sensor_controller:
    type:         force_torque_sensor_controller/ForceTorqueSensorController
-   publish_rate: 500
+   publish_rate: *loop_hz
 
 # Publish speed_scaling factor
 speed_scaling_state_controller:
    type:         ur_controllers/SpeedScalingStateController
-   publish_rate: 500
+   publish_rate: *loop_hz
 
 # Joint Trajectory Controller - position based -------------------------------
 # For detailed explanations of parameter see http://wiki.ros.org/joint_trajectory_controller
@@ -42,7 +42,7 @@ scaled_pos_traj_controller:
       wrist_2_joint: {trajectory: 0.2, goal: 0.1}
       wrist_3_joint: {trajectory: 0.2, goal: 0.1}
    stop_trajectory_duration: 0.5
-   state_publish_rate: 500
+   state_publish_rate: *loop_hz
    action_monitor_rate: 10
 
 pos_traj_controller:
@@ -58,5 +58,84 @@ pos_traj_controller:
       wrist_2_joint: {trajectory: 0.2, goal: 0.1}
       wrist_3_joint: {trajectory: 0.2, goal: 0.1}
    stop_trajectory_duration: 0.5
-   state_publish_rate: 500
+   state_publish_rate: *loop_hz
    action_monitor_rate: 10
+
+scaled_vel_traj_controller:
+   type: velocity_controllers/ScaledJointTrajectoryController
+   joints: *robot_joints
+   constraints:
+      goal_time: 0.6
+      stopped_velocity_tolerance: 0.05
+      shoulder_pan_joint: {trajectory: 0.1, goal: 0.1}
+      shoulder_lift_joint: {trajectory: 0.1, goal: 0.1}
+      elbow_joint: {trajectory: 0.1, goal: 0.1}
+      wrist_1_joint: {trajectory: 0.1, goal: 0.1}
+      wrist_2_joint: {trajectory: 0.1, goal: 0.1}
+      wrist_3_joint: {trajectory: 0.1, goal: 0.1}
+   stop_trajectory_duration: 0.5
+   state_publish_rate:  *loop_hz
+   action_monitor_rate: 10
+   gains:
+      #!!These values have not been optimized!!
+      shoulder_pan_joint:  {p: 5.0,  i: 0.05, d: 0.1, i_clamp: 1}
+      shoulder_lift_joint: {p: 5.0,  i: 0.05, d: 0.1, i_clamp: 1}
+      elbow_joint:         {p: 5.0,  i: 0.05, d: 0.1, i_clamp: 1}
+      wrist_1_joint:       {p: 5.0,  i: 0.05, d: 0.1, i_clamp: 1}
+      wrist_2_joint:       {p: 5.0,  i: 0.05, d: 0.1, i_clamp: 1}
+      wrist_3_joint:       {p: 5.0,  i: 0.05, d: 0.1, i_clamp: 1}
+
+   # Use a feedforward term to reduce the size of PID gains
+   velocity_ff:
+      shoulder_pan_joint: 1.0
+      shoulder_lift_joint: 1.0
+      elbow_joint: 1.0
+      wrist_1_joint: 1.0
+      wrist_2_joint: 1.0
+      wrist_3_joint: 1.0
+
+   # state_publish_rate:  50 # Defaults to 50
+   # action_monitor_rate: 20 # Defaults to 20
+   #stop_trajectory_duration: 0 # Defaults to 0.0
+
+vel_traj_controller:
+   type: velocity_controllers/JointTrajectoryController
+   joints: *robot_joints
+   constraints:
+      goal_time: 0.6
+      stopped_velocity_tolerance: 0.05
+      shoulder_pan_joint: {trajectory: 0.1, goal: 0.1}
+      shoulder_lift_joint: {trajectory: 0.1, goal: 0.1}
+      elbow_joint: {trajectory: 0.1, goal: 0.1}
+      wrist_1_joint: {trajectory: 0.1, goal: 0.1}
+      wrist_2_joint: {trajectory: 0.1, goal: 0.1}
+      wrist_3_joint: {trajectory: 0.1, goal: 0.1}
+   stop_trajectory_duration: 0.5
+   state_publish_rate:  *loop_hz
+   action_monitor_rate: 10
+   gains:
+      #!!These values have not been optimized!!
+      shoulder_pan_joint:  {p: 5.0,  i: 0.05, d: 0.1, i_clamp: 1}
+      shoulder_lift_joint: {p: 5.0,  i: 0.05, d: 0.1, i_clamp: 1}
+      elbow_joint:         {p: 5.0,  i: 0.05, d: 0.1, i_clamp: 1}
+      wrist_1_joint:       {p: 5.0,  i: 0.05, d: 0.1, i_clamp: 1}
+      wrist_2_joint:       {p: 5.0,  i: 0.05, d: 0.1, i_clamp: 1}
+      wrist_3_joint:       {p: 5.0,  i: 0.05, d: 0.1, i_clamp: 1}
+
+   # Use a feedforward term to reduce the size of PID gains
+   velocity_ff:
+      shoulder_pan_joint: 1.0
+      shoulder_lift_joint: 1.0
+      elbow_joint: 1.0
+      wrist_1_joint: 1.0
+      wrist_2_joint: 1.0
+      wrist_3_joint: 1.0
+
+   # state_publish_rate:  50 # Defaults to 50
+   # action_monitor_rate: 20 # Defaults to 20
+   #stop_trajectory_duration: 0 # Defaults to 0.0
+
+# Pass an array of joint velocities directly to the joints
+joint_group_vel_controller:
+   type: velocity_controllers/JointGroupVelocityController
+   joints: *robot_joints
--- a/ur_robot_driver/config/ur3e_controllers.yaml
+++ b/ur_robot_driver/config/ur3e_controllers.yaml
@@ -141,6 +141,6 @@ joint_group_vel_controller:
    joints: *robot_joints
 
 robot_status_controller:
-  type: industrial_robot_status_controller/IndustrialRobotStatusController
-  handle_name: industrial_robot_status_handle
-  publish_rate: 10
+   type: industrial_robot_status_controller/IndustrialRobotStatusController
+   handle_name: industrial_robot_status_handle
+   publish_rate: 10
--- a/ur_robot_driver/config/ur3e_controllers.yaml
+++ b/ur_robot_driver/config/ur3e_controllers.yaml
@@ -139,3 +139,8 @@ vel_traj_controller:
 joint_group_vel_controller:
    type: velocity_controllers/JointGroupVelocityController
    joints: *robot_joints
+
+robot_status_controller:
+  type: industrial_robot_status_controller/IndustrialRobotStatusController
+  handle_name: industrial_robot_status_handle
+  publish_rate: 10
--- a/ur_robot_driver/include/ur_robot_driver/rtde/rtde_client.h
+++ b/ur_robot_driver/include/ur_robot_driver/rtde/rtde_client.h
@@ -48,7 +48,6 @@ namespace ur_driver
 {
 namespace rtde_interface
 {
-
 enum UrRtdeRobotStatusBits
 {
   IS_POWER_ON = 0,
--- a/ur_robot_driver/include/ur_robot_driver/rtde/rtde_client.h
+++ b/ur_robot_driver/include/ur_robot_driver/rtde/rtde_client.h
@@ -48,7 +48,7 @@ namespace ur_driver
 {
 namespace rtde_interface
 {
-enum UrRtdeRobotStatusBits
+enum class UrRtdeRobotStatusBits
 {
   IS_POWER_ON = 0,
   IS_PROGRAM_RUNNING = 1,
@@ -56,7 +56,7 @@ enum UrRtdeRobotStatusBits
   IS_POWER_BUTTON_PRESSED = 3
 };
 
-enum UrRtdeSafetyStatusBits
+enum class UrRtdeSafetyStatusBits
 {
   IS_NORMAL_MODE = 0,
   IS_REDUCED_MODE = 1,
--- a/ur_robot_driver/include/ur_robot_driver/rtde/rtde_client.h
+++ b/ur_robot_driver/include/ur_robot_driver/rtde/rtde_client.h
@@ -48,6 +48,30 @@ namespace ur_driver
 {
 namespace rtde_interface
 {
+
+enum UrRtdeRobotStatusBits
+{
+  IS_POWER_ON = 0,
+  IS_PROGRAM_RUNNING = 1,
+  IS_TEACH_BUTTON_PRESSED = 2,
+  IS_POWER_BUTTON_PRESSED = 3
+};
+
+enum UrRtdeSafetyStatusBits
+{
+  IS_NORMAL_MODE = 0,
+  IS_REDUCED_MODE = 1,
+  IS_PROTECTIVE_STOPPED = 2,
+  IS_RECOVERY_MODE = 3,
+  IS_SAFEGUARD_STOPPED = 4,
+  IS_SYSTEM_EMERGENCY_STOPPED = 5,
+  IS_ROBOT_EMERGENCY_STOPPED = 6,
+  IS_EMERGENCY_STOPPED = 7,
+  IS_VIOLATION = 8,
+  IS_FAULT = 9,
+  IS_STOPPED_DUE_TO_SAFETY = 10
+};
+
 /*!
  * \brief The RTDEClient class manages communication over the RTDE interface. It contains the RTDE
  * handshake and read and write functionality to and from the robot.
--- a/ur_robot_driver/package.xml
+++ b/ur_robot_driver/package.xml
@@ -21,6 +21,7 @@
   <buildtool_depend>catkin</buildtool_depend>
 
   <build_depend>boost</build_depend>
+  <build_depend>industrial_robot_status_interface</build_depend>
 
   <depend>actionlib</depend>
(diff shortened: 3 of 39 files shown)


gavanderhoorn:
@axelschroth wrote in https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/issues/32#issuecomment-603310878:
> Does anyone have an idea, why no further state messages are published by industrial_robot_status_controller as soon as I stop the robot program (with ExternalControl Cap, see PR)?

The controller will only publish something if its `update(..)` is called.

You may have to add it to the `consistent_controllers` list [here](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/2ee8f2b1d1e71cae822bd8a488884d3b1907d317/ur_robot_driver/launch/ur_control.launch#L75-L79) so it doesn't get stopped.

ur_robot_driver/package.xml
```diff
@@ -38,6 +38,7 @@
   <depend>ur_dashboard_msgs</depend>
   <depend>ur_msgs</depend>
   <depend>std_srvs</depend>
+  <depend>industrial_robot_status_interface</depend>
```
gavanderhoorn: Should this be a `build_depend` only?

It's a header-only library, there is nothing at runtime there.

gavanderhoorn:
Would you not need to add `industrial_robot_status_interface` to the `COMPONENTS` of Catkin in the `CMakeLists.txt`?

ur_robot_driver/package.xml
axelschroth: Yes, you're right

axelschroth:
> Would you not need to add `industrial_robot_status_interface` to the `COMPONENTS` of Catkin in the `CMakeLists.txt`?

Done with last commit. Thanks! Any further suggestions?

gavanderhoorn:
Not right now.

Seeing as CI is 'down' at the moment (#144) and I'm not an maintainer here we'll have to wait for @fmauch to process the PR :)

But thanks for implementing this. :+1:

ur_robot_driver/src/ros/hardware_interface.cpp
```diff
@@ -628,6 +648,35 @@ void HardwareInterface::publishPose()
   }
 }
 
+void HardwareInterface::extractRobotStatus()
+{
+  // robot status bit 2: Is teach button pressed
+  robot_status_resource_.mode = robot_status_bits_[2] ? RobotMode::MANUAL : RobotMode::AUTO;
```
gavanderhoorn: I guess I do have an additional comment: please use some constants here for the bit indices instead of magic nrs.

fmauch: Is this really the right check? The robot can be in manual mode without the teach button being pressed, right?

I am aware, that usually on industrial robots manual and automatic modes have a different meaning than the one used on UR robots. But boiling it down to the teach button pressed feels misleading to me, as well.

gavanderhoorn: @fmauch: this was ported from `ur_modern_driver` I believe.

Would you happen to have a suggestion on how to better determine whether the controller is in `AUTO` or `MANUAL`?

I don't even believe URs really have that concept.

When they're being "freedrived" though they are definitely not in `AUTO` (or whatever we want to call the equivalent on URs).

fmauch: > I don't even believe URs really have that concept.

They do (on e-Series, on CB3 the same concept was called "running mode" and "programming mode"), but that concept is different (or a little less restrictive). In `AUTO` mode (which you can activate after setting a password) one can only load and execute predefined programs. (See section "Operational Mode Selection")

However, I am currently not aware of a method receiving this information. (which, of course, doesn't necessarily mean, that it doesn't exist)

> When they're being "freedrived" though they are definitely not in AUTO (or whatever we want to call the equivalent on URs).

Yes, but it feels as if we would be misusing the field from the message for something semantically different. Which, as you reminded me often, not the best thing to do :-)

fmauch: > Would you happen to have a suggestion on how to better determine whether the controller is in AUTO or MANUAL?

Forgot to answer that... To be honest - I don't know. I think, using what UR calls manual and automatic operational mode comes closest to what other industrial robots semantically state with this.

gavanderhoorn: > on e-Series, on CB3 the same concept was called "running mode" and "programming mode"

O yes, I'd forgotten about those. I hardly ever use the TP to program anything on these robots, hence why I hadn't thought of it.

> However, I am currently not aware of a method receiving this information. (which, of course, doesn't necessarily mean, that it doesn't exist)

Yes, would be great if we could get that information and use it here, as:

> it feels as if we would be misusing the field from the message for something semantically different. Which, as you reminded me often, not the best thing to do :-)

I completely agree here.

axelschroth: I agree with your points, but as you said @gavanderhoorn , this implementation is based on the interpretation of the ur_modern_driver.
I also thought about the issue when porting to the RTDE interface, but I didn't find any possibility to read this information about the Operational Mode over the RTDE interface (see [RTDE Guide](https://www.universal-robots.com/how-tos-and-faqs/how-to/ur-how-tos/real-time-data-exchange-rtde-guide/) )...

Another point to think of: What would we do with the CB-series? I think in the old Polyscope versions, you have to use a configurable input to change to running mode. Yet we could use that as "robot mode" if we have the information.

gavanderhoorn: What about limiting this PR to porting the `ur_modern_driver` implementation, then using a potential follow-up PR after/if we get more information on what @fmauch mentions?

It would be a nice enhancement over the partial implementation from `ur_modern_driver`.

@axelschroth: if you could address my comment about magic nrs which started this thread, I'd vote for merging it then.

axelschroth: I'd be happy about that, I will commit the changes about the magic nrs tomorrow.

fmauch: > What about limiting this PR to porting the `ur_modern_driver` implementation, then using a potential follow-up PR after/if we get more information on what @fmauch mentions?

I'd be fine with that.

axelschroth: I addressed your suggestion using an enum in the rtde client, because this belongs to the rtde specification. Are you okay with that?

ur_robot_driver/src/ros/hardware_interface.cpp
```diff
@@ -628,6 +648,38 @@ void HardwareInterface::publishPose()
...
+
+  robot_status_resource_.motion_possible =
+      robot_mode_ == ur_dashboard_msgs::RobotMode::RUNNING ? TriState::TRUE : TriState::FALSE;
+
+  // I found no way to reliably get information if the robot is moving
+  robot_status_resource_.in_motion = TriState::UNKNOWN;
+
+  // note that e-stop is handled by a seperate variable
+  robot_status_resource_.in_error =
+      safety_status_bits_[UrRtdeSafetyStatusBits::IS_PROTECTIVE_STOPPED] ? TriState::TRUE : TriState::FALSE;
```
gavanderhoorn: One thing I believe we could include in this PR: `in_error` should be the union of anything which makes the controller go into an error state. Is `IS_EMERGENCY_STOPPED` also a cause of `IS_PROTECTIVE_STOPPED`? If not, then it would need to be `||` here I believe.

Any other causes we could add to `in_error`?

axelschroth: Sure, I can add this. Unfortunately I cannot test emergency stops, since I don't have a real robot here, just ursim. I don't know the behaviour of these bits, therefore I cannot tell which bits provide additional information regarding the error state.

ur_robot_driver/include/ur_robot_driver/rtde/rtde_client.h
```diff
@@ -48,6 +48,29 @@ namespace ur_driver
 {
 namespace rtde_interface
 {
+enum UrRtdeRobotStatusBits
```
fmauch: Please make this scoped enums as we use this all over the driver.

ur_robot_driver/include/ur_robot_driver/rtde/rtde_client.h
```diff
@@ -48,6 +48,29 @@ namespace ur_driver
...
+{
+  IS_POWER_ON = 0,
+  IS_PROGRAM_RUNNING = 1,
+  IS_TEACH_BUTTON_PRESSED = 2,
+  IS_POWER_BUTTON_PRESSED = 3
+};
+
+enum UrRtdeSafetyStatusBits
```
fmauch: Scoped enums, same as above

ur_robot_driver/config/ur3e_controllers.yaml
```diff
@@ -139,3 +139,8 @@ vel_traj_controller:
 joint_group_vel_controller:
    type: velocity_controllers/JointGroupVelocityController
    joints: *robot_joints
+
+robot_status_controller:
+   type: industrial_robot_status_controller/IndustrialRobotStatusController
+   handle_name: industrial_robot_status_handle
+   publish_rate: 10
```
fmauch: This should be added for the other controller files, as well.

ur_robot_driver/src/ros/hardware_interface.cpp
fmauch: I think all of these should be used for that:
```
  IS_PROTECTIVE_STOPPED = 2,
  IS_ROBOT_EMERGENCY_STOPPED = 6,
  IS_EMERGENCY_STOPPED = 7,
  IS_VIOLATION = 8,
  IS_FAULT = 9,
  IS_STOPPED_DUE_TO_SAFETY = 10
```

fmauch: I could make a short trip to the lab these days to test the behavior.

ur_robot_driver/include/ur_robot_driver/rtde/rtde_client.h
axelschroth: Ok, then I'd rather switch to constants like

`static const uint8_t ROBOT_STATUS_BIT_IS_POWER_ON = 0;`

and so on, because accessing the bitsets with scoped enums gets quite ugly..

gavanderhoorn: I believe @fmauch already added a `std::underlying_type<>` somewhere?

ur_robot_driver/config/ur3e_controllers.yaml
axelschroth: Good point! Thanks for reviewing

gavanderhoorn:
Thanks for iterating @axelschroth :+1:

ur_robot_driver/src/ros/hardware_interface.cpp
```diff
@@ -628,6 +648,50 @@ void HardwareInterface::publishPose()
...
+
+  // I found no way to reliably get information if the robot is moving
+  robot_status_resource_.in_motion = TriState::UNKNOWN;
+
+  if (safety_status_bits_[toUnderlying(UrRtdeSafetyStatusBits::IS_PROTECTIVE_STOPPED)] ||
+      safety_status_bits_[toUnderlying(UrRtdeSafetyStatusBits::IS_ROBOT_EMERGENCY_STOPPED)] ||
+      safety_status_bits_[toUnderlying(UrRtdeSafetyStatusBits::IS_EMERGENCY_STOPPED)] ||
+      safety_status_bits_[toUnderlying(UrRtdeSafetyStatusBits::IS_VIOLATION)] ||
+      safety_status_bits_[toUnderlying(UrRtdeSafetyStatusBits::IS_FAULT)] ||
+      safety_status_bits_[toUnderlying(UrRtdeSafetyStatusBits::IS_STOPPED_DUE_TO_SAFETY)])
```
gavanderhoorn: This may be taking it too far, but: seeing as this set never changes, what about keeping a "mask" `bitset` around, initialised once and then used as:

```c++
if (safety_status_bits_ | in_error_bitset_)
   ...
```

`std::bitset` supports [the necessary operators](https://en.cppreference.com/w/cpp/utility/bitset/operator_logic2).

fmauch: I'll leave that up to @axelschroth It is fine like that from my side.

gavanderhoorn: We could ignore it in this PR, but seeing as this is run @ the controller's update rate (ie: 125 or 500 Hz), it's 7 operations vs 1. I'm not sure the compiler will be able to coalesce all the `bitset::test(..)` invocations into a single masking operation.

fmauch: true... And it should be fairly quick to implement :-)

fmauch:
Thanks again @axelschroth I well test that in the lab ASAP

axelschroth:
Thank you guys for reviewing @gavanderhoorn and @fmauch!

fmauch:
Testing results:

Robot running:
```
mode:
  val: 2
e_stopped:
  val: 0
drives_powered:
  val: 1
motion_possible:
  val: 1
in_motion:
  val: -1
in_error:
  val: 1
error_code: 0
```

EM-Stop hit:
```
mode:
  val: 2
e_stopped:
  val: 1
drives_powered:
  val: 1
motion_possible:
  val: 0
in_motion:
  val: -1
in_error:
  val: 1
error_code: 0
```

EM-Stop released:
```
mode:
  val: 2
e_stopped:
  val: 0
drives_powered:
  val: 1
motion_possible:
  val: 0
in_motion:
  val: -1
in_error:
  val: 1
error_code: 0
```

Protective stop:
```
mode:
  val: 2
e_stopped:
  val: 0
drives_powered:
  val: 1
motion_possible:
  val: 1
in_motion:
  val: -1
in_error:
  val: 1
error_code: 0
```

After Bootup:
```
mode:
  val: 2
e_stopped:
  val: 0
drives_powered:
  val: 0
motion_possible:
  val: 0
in_motion:
  val: -1
in_error:
  val: 1
error_code: 0
```

Switched on after bootup (not yet started):
```
mode:
  val: 2
e_stopped:
  val: 0
drives_powered:
  val: 1
motion_possible:
  val: 0
in_motion:
  val: -1
in_error:
  val: 1
error_code: 0
```

`in_error` seems to be wrong, I will look into that, later.

ur_robot_driver/src/ros/hardware_interface.cpp
```diff
@@ -628,6 +657,45 @@ void HardwareInterface::publishPose()
...
+  robot_status_resource_.drives_powered =
+      robot_status_bits_[toUnderlying(UrRtdeRobotStatusBits::IS_POWER_ON)] ? TriState::TRUE : TriState::FALSE;
+
+  robot_status_resource_.motion_possible =
+      robot_mode_ == ur_dashboard_msgs::RobotMode::RUNNING ? TriState::TRUE : TriState::FALSE;
+
+  // I found no way to reliably get information if the robot is moving
+  robot_status_resource_.in_motion = TriState::UNKNOWN;
+
+  if ((safety_status_bits_ | in_error_bitset_).any())
```
fmauch, line 685:
```suggestion
  if ((safety_status_bits_ & in_error_bitset_).any())
```

fmauch: yep, that fixes `in_error`

fmauch:
`motion_possible` in protective stop seems wrong, right?

ur_robot_driver/src/ros/hardware_interface.cpp
```diff
@@ -628,6 +657,45 @@ void HardwareInterface::publishPose()
...
+  robot_status_resource_.mode = robot_status_bits_[toUnderlying(UrRtdeRobotStatusBits::IS_TEACH_BUTTON_PRESSED)] ?
+                                    RobotMode::MANUAL :
+                                    RobotMode::AUTO;
+
+  robot_status_resource_.e_stopped = safety_status_bits_[toUnderlying(UrRtdeSafetyStatusBits::IS_EMERGENCY_STOPPED)] ?
+                                         TriState::TRUE :
+                                         TriState::FALSE;
+
+  // Note that this is true as soon as the drives are powered,
+  // even if the breakes are still closed
```
fmauch, line 673:
```suggestion
  // even if the brakes are still closed
```

ur_robot_driver/src/ros/hardware_interface.cpp
```diff
@@ -628,6 +657,45 @@ void HardwareInterface::publishPose()
...
+  {
+    robot_status_resource_.in_error = TriState::TRUE;
+  }
+  else
+  {
+    robot_status_resource_.in_error = TriState::FALSE;
+  }
+
+  // the error code, if any, is not transmitted by this protocol
+  // it can and should be fetched seperately
```
fmauch, line 695:
```suggestion
  // it can and should be fetched separately
```

ur_robot_driver/src/ros/hardware_interface.cpp
axelschroth: makes totally sense, thank you for testing and fixing!

ur_robot_driver/src/ros/hardware_interface.cpp
```diff
@@ -628,6 +657,45 @@ void HardwareInterface::publishPose()
...
+                                         TriState::FALSE;
+
+  // Note that this is true as soon as the drives are powered,
+  // even if the brakes are still closed
+  // which is in slight contrast to the comments in the
+  // message definition
+  robot_status_resource_.drives_powered =
+      robot_status_bits_[toUnderlying(UrRtdeRobotStatusBits::IS_POWER_ON)] ? TriState::TRUE : TriState::FALSE;
+
+  robot_status_resource_.motion_possible =
```
fmauch, line 679: I think a union would be good here, as well. Best achieved by combination with the following safety status bits:
```
  IS_PROTECTIVE_STOPPED = 2,
  IS_RECOVERY_MODE = 3,
  IS_SAFEGUARD_STOPPED = 4,
  IS_SYSTEM_EMERGENCY_STOPPED = 5,
  IS_ROBOT_EMERGENCY_STOPPED = 6,
  IS_EMERGENCY_STOPPED = 7,
  IS_VIOLATION = 8,
  IS_FAULT = 9,
  IS_STOPPED_DUE_TO_SAFETY = 10
```

Sorry for coming with that this late, it just came up while testing.

gavanderhoorn: Can't we check `in_error`? Are there any bits in this list which are not in the other?

fmauch: I did not propose `IS_SAFEGUARD_STOPPED` for `in_error` as I wouldn't consider that an error status. However, motion will not be possible, while this is active...

gavanderhoorn: Hm. Other industrial robots will report this as an error I believe.

If we don't, then we can run into the situation where `in_error` is `false` and `motors_on` is `true`, but `motion_possible` is `false`. That would at least be confusing.

fmauch: I thought, this would be exactly the purpose of this field. Otherwise it would be redundant, right?

But if that's the way this is usually being handled, I am absolutely fine with it.

gavanderhoorn: Anything which is not an error but would prevent motion should be used to determine the value of `in_motion`.

The two are not necessarily identical. Consider fi a Fanuc/KUKA/ABB/etc: controller in manual mode, TP in the hands of the operator. Operator holds the deadman switch(es), resets the controller, but does not turn on power -> `motion_possible` == `false`.

Enables drive power: `motion_possible` == `true`.

Keeps drive power enabled, but lets go of the deadman: `motion_possible` == `false`.

And in case there is a driver involved: no errors on the controller, drives powered, but the motion task of the driver is not active: `motion_possible` == `false`.

fmauch: This is exactly how I understood that field. And this is exactly my argument here, as well. I consider the `IS_SAFEGUARD_STOPPED` not an error. From my understanding this is used when an external device is connected to the controller that prevents motion once triggered, e.g. a door sensor, a laser scanner or a floor matt.

gavanderhoorn: Then `motion_possible` should be (pseudo):

```python
in_motion = in_error or (IS_SAFEGUARD_STOPPED in robot_status_bits_)
```

I believe.

axelschroth: Alright, I will modify the expression for motion_possible shortly. However I don't have time to include the status of the external control program, this condition may be added later if that's okay.

fmauch: Yes, absolutely!

axelschroth: > ```python
> motion_possible = in_error or (IS_SAFEGUARD_STOPPED in robot_status_bits_)
> ```
> 

I guess you mean the negated version of this expression. Motion is **not** possible if controller is in error or safeguard stop

gavanderhoorn: Yes, of course.

Updated.

fmauch:
I just tested this PR again and noticed one more issue: `robot_motion` always states `true` when there is no error and no safeguard stop active, even if the robot hardware is powered down completely. I made a PR to your branch @axelschroth with how I think the behavior is more correct.

axelschroth:
> I just tested this PR again and noticed one more issue: `robot_motion` always states `true` when there is no error and no safeguard stop active, even if the robot hardware is powered down completely. I made a PR to your branch @axelschroth with how I think the behavior is more correct.

Great, thank you! I merged your PR.

fmauch:
Perfect, I'll give it another HW test tomorrow and then merge it in if everything is fine. Thanks again for your continuous effors @axelschroth

gavanderhoorn:
Nice :tada:

Issue #146, opened 2020-03-26
No effort feedback


gtatiya:
# Summary

Hi, I'm trying to get effort feedback from the UR5 robot, but I'm not getting it. I'm running this commands:

Terminal 1:
`roslaunch ur_robot_driver ur5_bringup.launch robot_ip:=172.22.22.2 kinematics_config:="$(rospack find bimur_ur_launch)/etc/bimur_right_arm_calibration.yaml"`

Terminal 2:
`rostopic echo /joint_states`

## Actual Behavior
Here's the output:
![image](https://user-images.githubusercontent.com/22827159/77654689-c19eea80-6f47-11ea-98d7-058c38ed3a93.png)

As you can see effort values are zeros. It's zero even if I move the arm.

I followed the instructions provided to set the driver up and it works as I'm able to control the arm using `rosrun rqt_joint_trajectory_controller rqt_joint_trajectory_controller`.

I was getting effort values using the old driver which is deprecated: https://github.com/ros-industrial/ur_modern_driver.git.
## Expected Behavior
Here's the output of `rostopic echo /joint_states` for old driver:
![image](https://user-images.githubusercontent.com/22827159/77654712-c8c5f880-6f47-11ea-84db-3a180f10c0cd.png)

How can get the effort feedback?

# Versions
 - ROS Driver version: Kinetic
 - Affected Robot Software Version(s): URSoftware 3.7
 - Robot Serial Number: 2018359497
 - UR+ product(s) installed: UR5
 - URCaps Software version(s): External Control 1.0.1

fmauch:
Yes, this was never implemented.

The reasoning for this is that the robot does not provide any effort information. The `ur_modern_driver` was publishing the joint currents to the effort field instead.

This being said, if you're only interested in getting the joint currents, this will be resolved in #35. If you would vote for having the joint currents in the `effort` field of the `joint_state` we can discuss that here :-)

gavanderhoorn:
@fmauch wrote:
> The reasoning for this is that the robot does not provide any effort information.

well, there is `get_joint_torques()`.

Not sure whether we can get the same information into the RTDE output?

fmauch:
Well, there is `target_moment` but no `actual` field.

gavanderhoorn:
Hm.

Could be worth a check with UR?

fmauch:
> Hm.
> 
> Could be worth a check with UR?

Definitely

> 
[...]

PR #160, opened 2020-04-16
get effort feedback in joint_states

gtatiya:
@fmauch I made all the changes according to [link](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/issues/146#issuecomment-610186709).

@gavanderhoorn I forked the repo. and made a pull request here.

Here are the changes:
- [ros_control.urscript](https://github.com/gtatiya/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/resources/ros_control.urscript)

- [rtde_output_recipe.txt](https://github.com/gtatiya/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/resources/rtde_output_recipe.txt)

- [Read value](https://github.com/gtatiya/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/src/ros/hardware_interface.cpp#L416-L421)

- [Store it inside a member declared](https://github.com/gtatiya/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/include/ur_robot_driver/ros/hardware_interface.h#L245)

- [Publish](https://github.com/gtatiya/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/src/ros/hardware_interface.cpp#L714-L719)

I'm not sure if I should use `output_double_register_24` here: `tool_data_pub_->msg_.output_double_register_24 = output_double_register_[0];`

When I run `roslaunch ur_robot_driver ur5_bringup.launch robot_ip:=127.0.0.1`, I get this error:
image

I guess, this error is raised [here](https://github.com/gtatiya/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/src/ros/hardware_interface.cpp#L483)

`rosrun rqt_joint_trajectory_controller rqt_joint_trajectory_controller` gives this error:
image

How to fix this error?
How to get the effort feedback in `joint_states`?

In addition, `output_double_register_X` was added for `software version >=3.9.0`: https://www.universal-robots.com/articles/ur-articles/real-time-data-exchange-rtde-guide/
I was working on 3.7, so I had to upgrade. Thus, if everything work, you need to update the [README](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/master/README.md) to `software version >=3.9.0`.

### DIFF
--- a/ur_robot_driver/include/ur_robot_driver/ros/hardware_interface.h
+++ b/ur_robot_driver/include/ur_robot_driver/ros/hardware_interface.h
@@ -252,7 +252,8 @@ protected:
   int32_t safety_mode_;
   std::bitset<4> robot_status_bits_;
   std::bitset<11> safety_status_bits_;
-
+  std::array<double, 6> output_double_register_;
+  
   std::unique_ptr<realtime_tools::RealtimePublisher<tf2_msgs::TFMessage>> tcp_pose_pub_;
   std::unique_ptr<realtime_tools::RealtimePublisher<ur_msgs::IOStates>> io_pub_;
   std::unique_ptr<realtime_tools::RealtimePublisher<ur_msgs::ToolDataMsg>> tool_data_pub_;
--- a/ur_robot_driver/include/ur_robot_driver/ros/hardware_interface.h
+++ b/ur_robot_driver/include/ur_robot_driver/ros/hardware_interface.h
@@ -252,7 +252,7 @@ protected:
   int32_t safety_mode_;
   std::bitset<4> robot_status_bits_;
   std::bitset<11> safety_status_bits_;
-  
+
   std::unique_ptr<realtime_tools::RealtimePublisher<tf2_msgs::TFMessage>> tcp_pose_pub_;
   std::unique_ptr<realtime_tools::RealtimePublisher<ur_msgs::IOStates>> io_pub_;
   std::unique_ptr<realtime_tools::RealtimePublisher<ur_msgs::ToolDataMsg>> tool_data_pub_;
--- a/ur_robot_driver/include/ur_robot_driver/ros/hardware_interface.h
+++ b/ur_robot_driver/include/ur_robot_driver/ros/hardware_interface.h
@@ -252,7 +252,6 @@ protected:
   int32_t safety_mode_;
   std::bitset<4> robot_status_bits_;
   std::bitset<11> safety_status_bits_;
-  std::array<double, 6> output_double_register_;
   
   std::unique_ptr<realtime_tools::RealtimePublisher<tf2_msgs::TFMessage>> tcp_pose_pub_;
   std::unique_ptr<realtime_tools::RealtimePublisher<ur_msgs::IOStates>> io_pub_;
--- a/ur_robot_driver/resources/rtde_output_recipe.txt
+++ b/ur_robot_driver/resources/rtde_output_recipe.txt
@@ -24,9 +24,4 @@ robot_mode
 safety_mode
 robot_status_bits
 safety_status_bits
-output_double_register_0
-output_double_register_1
-output_double_register_2
-output_double_register_3
-output_double_register_4
-output_double_register_5
+actual_current
--- a/ur_robot_driver/resources/rtde_output_recipe.txt
+++ b/ur_robot_driver/resources/rtde_output_recipe.txt
@@ -22,20 +22,11 @@ tool_output_current
 tool_temperature
 robot_mode
 safety_mode
-<<<<<<< HEAD
 robot_status_bits
 safety_status_bits
-output_double_register_24
-output_double_register_25
-output_double_register_26
-output_double_register_27
-output_double_register_28
-output_double_register_29
-=======
 output_double_register_0
 output_double_register_1
 output_double_register_2
 output_double_register_3
 output_double_register_4
 output_double_register_5
->>>>>>> changed register no. from 24-29 to 0-9 in rtde_output_recipe.txt
--- a/ur_robot_driver/resources/rtde_output_recipe.txt
+++ b/ur_robot_driver/resources/rtde_output_recipe.txt
@@ -22,6 +22,7 @@ tool_output_current
 tool_temperature
 robot_mode
 safety_mode
+<<<<<<< HEAD
 robot_status_bits
 safety_status_bits
 output_double_register_24
@@ -30,3 +31,11 @@ output_double_register_26
 output_double_register_27
 output_double_register_28
 output_double_register_29
+=======
+output_double_register_0
+output_double_register_1
+output_double_register_2
+output_double_register_3
+output_double_register_4
+output_double_register_5
+>>>>>>> changed register no. from 24-29 to 0-9 in rtde_output_recipe.txt
--- a/ur_robot_driver/resources/rtde_output_recipe.txt
+++ b/ur_robot_driver/resources/rtde_output_recipe.txt
@@ -24,3 +24,9 @@ robot_mode
 safety_mode
 robot_status_bits
 safety_status_bits
+output_double_register_24
+output_double_register_25
+output_double_register_26
+output_double_register_27
+output_double_register_28
+output_double_register_29
--- a/ur_robot_driver/src/ros/hardware_interface.cpp
+++ b/ur_robot_driver/src/ros/hardware_interface.cpp
@@ -442,6 +442,12 @@ void HardwareInterface::read(const ros::Time& time, const ros::Duration& period)
     readData(data_pkg, "safety_mode", safety_mode_);
     readBitsetData<uint32_t>(data_pkg, "robot_status_bits", robot_status_bits_);
     readBitsetData<uint32_t>(data_pkg, "safety_status_bits", safety_status_bits_);
+    readData(data_pkg, "output_double_register_24", output_double_register_[0]);
+    readData(data_pkg, "output_double_register_25", output_double_register_[1]);
+    readData(data_pkg, "output_double_register_26", output_double_register_[2]);
+    readData(data_pkg, "output_double_register_27", output_double_register_[3]);
+    readData(data_pkg, "output_double_register_28", output_double_register_[4]);
+    readData(data_pkg, "output_double_register_29", output_double_register_[5]);
     readBitsetData<uint64_t>(data_pkg, "actual_digital_input_bits", actual_dig_in_bits_);
     readBitsetData<uint64_t>(data_pkg, "actual_digital_output_bits", actual_dig_out_bits_);
     readBitsetData<uint32_t>(data_pkg, "analog_io_types", analog_io_types_);
@@ -789,6 +795,12 @@ void HardwareInterface::publishToolData()
       tool_data_pub_->msg_.tool_output_voltage = tool_output_voltage_;
       tool_data_pub_->msg_.tool_current = tool_output_current_;
       tool_data_pub_->msg_.tool_temperature = tool_temperature_;
+      tool_data_pub_->msg_.output_double_register_24 = output_double_register_[0];
+      tool_data_pub_->msg_.output_double_register_25 = output_double_register_[1];
+      tool_data_pub_->msg_.output_double_register_26 = output_double_register_[2];
+      tool_data_pub_->msg_.output_double_register_27 = output_double_register_[3];
+      tool_data_pub_->msg_.output_double_register_28 = output_double_register_[4];
+      tool_data_pub_->msg_.output_double_register_29 = output_double_register_[5];
       tool_data_pub_->unlockAndPublish();
     }
   }
--- a/ur_robot_driver/src/ros/hardware_interface.cpp
+++ b/ur_robot_driver/src/ros/hardware_interface.cpp
@@ -442,12 +442,12 @@ void HardwareInterface::read(const ros::Time& time, const ros::Duration& period)
     readData(data_pkg, "safety_mode", safety_mode_);
     readBitsetData<uint32_t>(data_pkg, "robot_status_bits", robot_status_bits_);
     readBitsetData<uint32_t>(data_pkg, "safety_status_bits", safety_status_bits_);
-    readData(data_pkg, "output_double_register_0", output_double_register_[0]);
-    readData(data_pkg, "output_double_register_1", output_double_register_[1]);
-    readData(data_pkg, "output_double_register_2", output_double_register_[2]);
-    readData(data_pkg, "output_double_register_3", output_double_register_[3]);
-    readData(data_pkg, "output_double_register_4", output_double_register_[4]);
-    readData(data_pkg, "output_double_register_5", output_double_register_[5]);
+    readData(data_pkg, "output_double_register_0", joint_efforts_[0]);
+    readData(data_pkg, "output_double_register_1", joint_efforts_[1]);
+    readData(data_pkg, "output_double_register_2", joint_efforts_[2]);
+    readData(data_pkg, "output_double_register_3", joint_efforts_[3]);
+    readData(data_pkg, "output_double_register_4", joint_efforts_[4]);
+    readData(data_pkg, "output_double_register_5", joint_efforts_[5]);
     readBitsetData<uint64_t>(data_pkg, "actual_digital_input_bits", actual_dig_in_bits_);
     readBitsetData<uint64_t>(data_pkg, "actual_digital_output_bits", actual_dig_out_bits_);
     readBitsetData<uint32_t>(data_pkg, "analog_io_types", analog_io_types_);
@@ -795,12 +795,6 @@ void HardwareInterface::publishToolData()
       tool_data_pub_->msg_.tool_output_voltage = tool_output_voltage_;
       tool_data_pub_->msg_.tool_current = tool_output_current_;
       tool_data_pub_->msg_.tool_temperature = tool_temperature_;
-      tool_data_pub_->msg_.output_double_register_24 = output_double_register_[0];
-      tool_data_pub_->msg_.output_double_register_25 = output_double_register_[1];
-      tool_data_pub_->msg_.output_double_register_26 = output_double_register_[2];
-      tool_data_pub_->msg_.output_double_register_27 = output_double_register_[3];
-      tool_data_pub_->msg_.output_double_register_28 = output_double_register_[4];
-      tool_data_pub_->msg_.output_double_register_29 = output_double_register_[5];
       tool_data_pub_->unlockAndPublish();
     }
   }
--- a/ur_robot_driver/src/ros/hardware_interface.cpp
+++ b/ur_robot_driver/src/ros/hardware_interface.cpp
@@ -440,18 +440,9 @@ void HardwareInterface::read(const ros::Time& time, const ros::Duration& period)
     readData(data_pkg, "tool_temperature", tool_temperature_);
     readData(data_pkg, "robot_mode", robot_mode_);
     readData(data_pkg, "safety_mode", safety_mode_);
-<<<<<<< HEAD
     readBitsetData<uint32_t>(data_pkg, "robot_status_bits", robot_status_bits_);
     readBitsetData<uint32_t>(data_pkg, "safety_status_bits", safety_status_bits_);
-    readData(data_pkg, "output_double_register_0", joint_efforts_[0]);
-    readData(data_pkg, "output_double_register_1", joint_efforts_[1]);
-    readData(data_pkg, "output_double_register_2", joint_efforts_[2]);
-    readData(data_pkg, "output_double_register_3", joint_efforts_[3]);
-    readData(data_pkg, "output_double_register_4", joint_efforts_[4]);
-    readData(data_pkg, "output_double_register_5", joint_efforts_[5]);
-=======
     readData(data_pkg, "actual_current", joint_efforts_);
->>>>>>> effort in joint_states now outputs current
     readBitsetData<uint64_t>(data_pkg, "actual_digital_input_bits", actual_dig_in_bits_);
     readBitsetData<uint64_t>(data_pkg, "actual_digital_output_bits", actual_dig_out_bits_);
     readBitsetData<uint32_t>(data_pkg, "analog_io_types", analog_io_types_);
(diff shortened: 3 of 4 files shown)


ur_robot_driver/src/ros/hardware_interface.cpp
```diff
@@ -705,6 +711,12 @@ void HardwareInterface::publishToolData()
       tool_data_pub_->msg_.tool_output_voltage = tool_output_voltage_;
       tool_data_pub_->msg_.tool_current = tool_output_current_;
       tool_data_pub_->msg_.tool_temperature = tool_temperature_;
+      tool_data_pub_->msg_.output_double_register_24 = output_double_register_[0];
+      tool_data_pub_->msg_.output_double_register_25 = output_double_register_[1];
+      tool_data_pub_->msg_.output_double_register_26 = output_double_register_[2];
+      tool_data_pub_->msg_.output_double_register_27 = output_double_register_[3];
+      tool_data_pub_->msg_.output_double_register_28 = output_double_register_[4];
+      tool_data_pub_->msg_.output_double_register_29 = output_double_register_[5];
```
fmauch, line 719: This does obviously not build, as the message doesn't contain such a field.

ur_robot_driver/resources/rtde_output_recipe.txt
```diff
@@ -22,3 +22,9 @@ tool_output_current
 tool_temperature
 robot_mode
 safety_mode
+output_double_register_24
+output_double_register_25
+output_double_register_26
+output_double_register_27
+output_double_register_28
+output_double_register_29
```
fmauch, line 30: Note that you wrote the torque data into registers 4-9, not 24-29. It might make sense to read those instead.

fmauch:
> In addition, output_double_register_X was added for software version >=3.9.0: https://www.universal-robots.com/articles/ur-articles/real-time-data-exchange-rtde-guide/
I was working on 3.7, so I had to upgrade. Thus, if everything work, you need to update the README to software version >=3.9.0.

Then I would vote for using something in the range of 0-23 instead.

ur_robot_driver/src/ros/hardware_interface.cpp
gavanderhoorn: I would also suggest to include the effort data in the appropriate message (ie: `JointState`). This can be accomplished by updating the appropriate member variable of the hw interface (which in the end gets read by the `joint_state_controller`).

fmauch: I think that was the plan initially.

ur_robot_driver/resources/rtde_output_recipe.txt
tonkei0361: @fmauch I implemented the changes above on my local `rtde_output_recipe.txt` but the error "Could not get fresh data package from robot" still appeared. After looking through URscript documentation and software requirements, I found that for e-series, the software version must be at least 5.3 to access registers 24..47. However, my robot is at 5.3.1 and I still see this error. If I keep the register index below 24, no errors show up.

Instead of outputting effort data to `output_double_register_X`, if I want to incorporate effort data into `JointState` message, how can I do that?

gtatiya:
@fmauch @gavanderhoorn I changed the register no. to 0 to 6 in `ros_control.urscript`, `rtde_output_recipe.txt` and `hardware_interface.cpp`, and pushed it: https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/160/commits/b64de6f45dc796ce0501a3011aa0f2474c579230.

But, I'm still not sure about [this](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/b64de6f45dc796ce0501a3011aa0f2474c579230/ur_robot_driver/src/ros/hardware_interface.cpp#L714-L719), so I kept it as it was.

I know that message type should be this: http://docs.ros.org/melodic/api/sensor_msgs/html/msg/JointState.html

But, I do not sure how to use it in `hardware_interface.cpp`.

Do I need to write a new function like `publishRTMsg()` in the old driver: https://github.com/ros-industrial/ur_modern_driver/blob/9397f47ac158fe5976719f33101334371bc15d44/src/ur_ros_wrapper.cpp?

gavanderhoorn:
@gtatiya wrote:
> But, I'm still not sure about [this](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/b64de6f45dc796ce0501a3011aa0f2474c579230/ur_robot_driver/src/ros/hardware_interface.cpp#L714-L719), so I kept it as it was.

That needs to be removed.

 See below.

> I know that message type should be this: http://docs.ros.org/melodic/api/sensor_msgs/html/msg/JointState.html
> 
> But, I do not sure how to use it in `hardware_interface.cpp`.
> 
> Do I need to write a new function like `publishRTMsg()` in the old driver: https://github.com/ros-industrial/ur_modern_driver/blob/9397f47ac158fe5976719f33101334371bc15d44/src/ur_ros_wrapper.cpp?

No, you don't need to create any new function.

All you need to do should be to update the [joint_efforts_](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/0baaa19bb16960c024b289dbaedd8f9357099c17/ur_robot_driver/include/ur_robot_driver/ros/hardware_interface.h#L222) member with the data encoded in the RTDE message.

For `joint_positions` and `joint_velocities`, that's done here:

https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/0baaa19bb16960c024b289dbaedd8f9357099c17/ur_robot_driver/src/ros/hardware_interface.cpp#L397-L398

You should be able to add something for `joint_efforts_` there, but you'll probably have to account for the fact that `joint_efforts_` is a single 6D vector, while you encoded the efforts in 6 separate float field in the message.

That's not necessarily a problem, but it does mean you cannot just use a single `readData(..)` invocation as is done for `joint_positions` and the other fields.

I haven't tested this, but perhaps you could do something like this:

```c++
readData(data_pkg, "output_double_register_24", joint_efforts_[0]);
readData(data_pkg, "output_double_register_25", joint_efforts_[1]);

..etc
```

As I wrote: I haven't tested this at all and it might be there is a more efficient way to do this, but I'm sure @fmauch will tell us if that is the case.

---

Updating `joint_efforts_` is all that is needed for the `joint_state_controller` to publish the efforts as part of the `JointState` message it already publishes.

ur_robot_driver/src/ros/hardware_interface.cpp
gavanderhoorn: @gtatiya: you should remove these lines.

ur_robot_driver/src/ros/hardware_interface.cpp
```diff
@@ -413,6 +413,12 @@ void HardwareInterface::read(const ros::Time& time, const ros::Duration& period)
     readData(data_pkg, "tool_temperature", tool_temperature_);
     readData(data_pkg, "robot_mode", robot_mode_);
     readData(data_pkg, "safety_mode", safety_mode_);
+    readData(data_pkg, "output_double_register_0", output_double_register_[0]);
+    readData(data_pkg, "output_double_register_1", output_double_register_[1]);
+    readData(data_pkg, "output_double_register_2", output_double_register_[2]);
+    readData(data_pkg, "output_double_register_3", output_double_register_[3]);
+    readData(data_pkg, "output_double_register_4", output_double_register_[4]);
+    readData(data_pkg, "output_double_register_5", output_double_register_[5]);
```
gavanderhoorn, line 421: See https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/160#issuecomment-615733315 for how you should adapt this to something which *may* work.

ur_robot_driver/include/ur_robot_driver/ros/hardware_interface.h
```diff
@@ -242,7 +242,8 @@ class HardwareInterface : public hardware_interface::RobotHW
   std::vector<std::string> joint_names_;
   int32_t robot_mode_;
   int32_t safety_mode_;
-
+  std::array<double, 6> output_double_register_;
+
```
gavanderhoorn, line 246: This line is not needed. See my comment. You should be able to just use `joint_efforts_`, which is already a member of this class.

gtatiya:
I made changes suggested by @gavanderhoorn and pushed: https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/160/commits/896d4437ac13d9626ce97b07072f7d8bfb4d02bc

Now, when I run `roslaunch ur_robot_driver ur5_bringup.launch robot_ip:=127.0.0.1` and I'm able to control the arm using `rosrun rqt_joint_trajectory_controller rqt_joint_trajectory_controller`.

However, when I run `rostopic echo /joint_states`, the effort values do not change even when I move the arm:
image

@gavanderhoorn @fmauch Could you please help me figure out what is going wrong?

gavanderhoorn:
Well, as I wrote: I did not test my suggestion, so it could be something is not quite as it should be.

I can't do any debugging myself at the moment, so we'll have to wait for @fmauch.

fmauch:
@gavanderhoorn absolutely did the right suggestions. 

I just tested this PR and it seems to work:
```
---
header:
  seq: 183082
  stamp:
    secs: 1587362923
    nsecs: 959769376
  frame_id: ''
name: - elbow_joint
- shoulder_lift_joint
- shoulder_pan_joint
- wrist_1_joint
- wrist_2_joint
- wrist_3_joint
position: [2.47056228319277, -2.2998558483519496, 2.6780805587768555, -2.314638753930563, 2.251518726348877, -0.750075642262594]
velocity: [-0.0, 0.0, 4.15845160921664e-16, 0.0, 0.0, 0.0]
effort: [3.1766276359558105, 2.8468430042266846, -3.1247429847717285, 2.3760974407196045, 4.327432155609131, -1.142044186592102]
```
As you are writing the effort data into the registers from the URScript part, your program containing that code must actually be running in order to run this. This means, you'll have to have a program running on the TP containing the "ExternalControl" node.

ur_robot_driver/include/ur_robot_driver/ros/hardware_interface.h
```diff
@@ -242,7 +242,7 @@ class HardwareInterface : public hardware_interface::RobotHW
   std::vector<std::string> joint_names_;
   int32_t robot_mode_;
   int32_t safety_mode_;
-
+
```
fmauch, line 245: Please do not introduce whitespace lines. This is probably the reason the format-check fails.

gtatiya:
@fmauch It's good that it works on the real robot.

> As you are writing the effort data into the registers from the URScript part, your program containing that code must actually be running in order to run this. This means, you'll have to have a program running on the TP containing the "ExternalControl" node.
> 
> BTW: URsim does indeed report 0 values here, so this is not testable on a simulated robot.

Yes, I tesed it on URSim because my University suggested to avoid going to the lab as much as possible, but I will go to test this soon.
Before I test it on the real robot, I have a few questions:

- Earlier [you mentioned](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/issues/146#issuecomment-610186709) that "When the driver and External Control are running, [script code](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/resources/ros_control.urscript) is being sent to the robot and executed there." So, why I need to run a seperate program to make it work?

- How to have a program running on the TP containing the "ExternalControl" node?

- What this program should contain?

- What is TP here?

Once I'm sure how to run it, I'll go to test it on the real robot, and report you.

ur_robot_driver/include/ur_robot_driver/ros/hardware_interface.h
gtatiya: I removed it, and I will keep that in mind.

fmauch:
> Before I test it on the real robot, I have a few questions:
> 
> * Earlier [you mentioned](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/issues/146#issuecomment-610186709) that "When the driver and External Control are running, [script code](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/resources/ros_control.urscript) is being sent to the robot and executed there." So, why I need to run a seperate program to make it work?

Becaus of the term "External Control are running". This means exactly that. You have to have a program running containing the "External Control" program node. Reasons for that can be found [here](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/tree/master/ur_robot_driver#technical-details).

> * How to have a program running on the TP containing the "ExternalControl" node?
> * What this program should contain?

See the [setup guide](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/doc/install_urcap_cb3.md)

> * What is TP here?

The teach pendant. The gui that you get from the robot.

gtatiya:
@fmauch

> Becaus of the term "External Control are running". This means exactly that. You have to have a program running containing the "External Control" program node. Reasons for that can be found [here](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/tree/master/ur_robot_driver#technical-details).

> See the [setup guide](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/doc/install_urcap_cb3.md)

I just read the documentation. I got confused because I always run the "External Control" program on the TP or the URSim as mentioned [here](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/doc/install_urcap_cb3.md). So, now I do not think I need to do anything extra.

I do have one more question:
You [suggested](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/160#issuecomment-615080074) to use `output_double_register_0` to `output_double_register_5` which is for FieldBus/PLC (I'm not sure what's that) as mentioned in [RTDE guide](https://www.universal-robots.com/articles/ur-articles/real-time-data-exchange-rtde-guide/), and registers `24..47` is reserved for `external RTDE clients (i.e URCAPS)`, so do you think if I use  `output_double_register_24` to `output_double_register_29` it will work on URSim?

fmauch:
I think, it will work with a version recent enough, but that's just a guess. But you are equipped with everything to test it yourself :-)

However, I would rather keep it at the first 24 registers in order to not raise the version bar for this driver.

gavanderhoorn:
If the RTDE guide suggests to use 24-47 for the kind of usage we're implementing here, it would perhaps be sensible to follow that suggesting.

I do understand @fmauch's reluctance though: it would mean increasing the required version of the system software.

gtatiya:
@fmauch @gavanderhoorn I tried using URSim 3.7.0 and 3.12.1 for both set of registers, and here's the results:

**URSim 3.7.0**
`output_double_register_0 to output_double_register_5`: Works

`output_double_register_24 to output_double_register_29`: Gives error:
image

**URSim 3.12.1**
`output_double_register_0 to output_double_register_5`: Works

`output_double_register_24 to output_double_register_29`: Gives error:
image

I have 3.7.0 installed on TP of the real robot, so I will keep `output_double_register_0 to output_double_register_5` and if it does not work, I will upgrade. I'm afraid to upgrade because TP does not support downgrading the software.

fmauch:
The result of your 3.12 test surprise me and should be further investigated (outside of this PR).

However, I would suggest that we add two parameters to the HW interface
 * `enable_efforts` that will opt in the changes suggested in this PR. This will need proper documentation though, to make it clear that this depends on the "External Control" program node being active to work.
 * `effort_register_offset` an option to customize the start of the six registers used for transferring efforts.

Then, there's still one more question popping up, maybe @gavanderhoorn can answer that: Is this the correct value to add to the `joint_state` efforts? From my understanding this reflects the "external" torques applied to the robot, as the efforts needed to move the robot around are calculated out of the equation. Intuitively I would expect that a `joint_state` effort shows me the actual efforts commanded to the robot, which we don't have available directly (we have the desired state `target_moment` and the `actual_current`, which should be proportional to the actual moments). Are we doing the right thing here?

gavanderhoorn:
> Is this the correct value to add to the `joint_state` efforts? From my understanding this reflects the "external" torques applied to the robot, as the efforts needed to move the robot around are calculated out of the equation. 

That would indeed be what `get_joint_torques()` returns:

> The torque on the joints, corrected by the torque needed to move therobot itself (gravity, friction, etc.)

That's from the script manual documentation on `get_joint_torques()`.

> Intuitively I would expect that a `joint_state` effort shows me the actual efforts commanded to the robot, which we don't have available directly (we have the desired state `target_moment` and the `actual_current`, which should be proportional to the actual moments). Are we doing the right thing here?

The documentation for [JointState](http://docs.ros.org/api/sensor_msgs/html/msg/JointState.html) (as little as there is) does indeed say that `effort` should contain the:

> the effort that is applied in the joint (Nm or N)

It would seem that what comes out of `get_joint_torques()` is not what we need.

Thinking back, I believe this was also the reason why `ur_modern_driver` publishes the current. It's not exactly what it should be, but it's at least the commanded effort (albeit not in Nm).

gtatiya:
@fmauch @gavanderhoorn I'm in the lab with the real robot. I tested the code, but I'm still getting 0s for effort.

Here's the steps I did:
- `roslaunch ur_robot_driver ur5_bringup.launch robot_ip:=172.22.22.2 kinematics_config:="$(rospack find bimur_ur_launch)/etc/bimur_right_arm_calibration.yaml"`

- Launched "External Control" program on TP with my PC's IP

- `rosrun rqt_joint_trajectory_controller rqt_joint_trajectory_controller`

- `rostopic echo /joint_states`

Here's the output:
image

I have UR Software 3.7.0 installed in the TP.

What could be wrong? Am I forgetting anything? I'll be the lab for a few hours, let me know what else I should try?

gtatiya:
@fmauch @gavanderhoorn Could you guys answer my last message.

gtatiya:
@tonkei0361 Did you tried the last update?

tonkei0361:
@gtatiya I tried the code from this commit and it worked, thanks! Originally, I was doing similar modifications and instead of using registers 0 to 5, I used 24-29.

gtatiya:
> @gtatiya I tried the code from this commit and it worked, thanks! Originally, I was doing similar modifications and instead of using registers 0 to 5, I used 24-29.

@tonkei0361 
What UR Software version you have installed on TP? I'm still not getting effort values, do you see issue with the steps I followed here: https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/160#issuecomment-618367150?

fmauch:
@gtatiya As this PR seems to work basically, there's not too much I can help you instead of suggesting the usual debugging steps like "Are you sure, you have built your workspace with the correct branch"? "Did you get any error output during startup"?

tonkei0361:
> > @gtatiya I tried the code from this commit and it worked, thanks! Originally, I was doing similar modifications and instead of using registers 0 to 5, I used 24-29.
> 
> @tonkei0361
> What UR Software version you have installed on TP? I'm still not getting effort values, do you see issue with the steps I followed here: [#160 (comment)](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/160#issuecomment-618367150)?

@gtatiya My robot is UR5e, so the TP software is different. Mine is 5.3.1 and I did exactly what you did on the terminal. Sorry that I can't help much:(

tonkei0361:
> The documentation for [JointState](http://docs.ros.org/api/sensor_msgs/html/msg/JointState.html) (as little as there is) does indeed say that `effort` should contain:
> 
> > the effort that is applied in the joint (Nm or N)
> 
> It would seem that what comes out of `get_joint_torques()` is not what we need.
> 
> Thinking back, I believe this was also the reason why `ur_modern_driver` publishes the current. It's not exactly what it should be, but it's at least the commanded effort (albeit not in Nm).

Based on the above comments from @gavanderhoorn , I modified my code so that `effort` now publishes the current from `actual_current` instead of returned values of `get_joint_torques()`. This makes more sense to me. Despite the unit mismatch as mentioned above, `get_joint_torques()` does not go back to zero after an external torque is applied to a joint. I think that this can be problematic when effort is used in a feedback loop.

@fmauch Is it possible to merge this pull request? Also, should I commit my changes here?

gtatiya:
> @fmauch Is it possible to merge this pull request? Also, should I commit my changes here?

@tonkei0361 If you like I can make you a collaborator to the fork I used, and then you can commit. Let me know what you think.

tonkei0361:
> > @fmauch Is it possible to merge this pull request? Also, should I commit my changes here?
> 
> @tonkei0361 If you like I can make you a collaborator to the fork I used, and then you can commit. Let me know what you think.

@gtatiya I would appreciate that! After I commit my changes, you can try it on your UR5 to see if you read any effort values. Mine works with UR5e.

gtatiya:
> @gtatiya I would appreciate that! After I commit my changes, you can try it on your UR5 to see if you read any effort values. Mine works with UR5e.

@tonkei0361 I added you. After you push, I will test it in a few days.

fmauch:
Thanks for sticking around with this. I still think, that publishing the actual joint torques is interesting to some people, so we should not simply throw away your efforts there, but for the `joint_state` efforts it might not be the right thing.

I'll give it another go testing and then will merge it.

tonkei0361:
> Thanks for sticking around with this. I still think, that publishing the actual joint torques is interesting to some people, so we should not simply throw away your efforts there, but for the `joint_state` efforts it might not be the right thing.

@fmauch What gets me confused is that the returned value of `get_joint_torques()` does not return to the same initial value after an external torque is applied to the joints. Let's say that when I initially start the driver, the torque from one joint is 0.3. When I apply a torque to that joint, the value changes, so far so good. As soon as I let go of the torque, the torque reading from that joint somehow doesn't return to 0.3. Do you know what is going on here?

We can indeed keep the method of outputting joint torques through output registers. If needed, I can add back the output registers and forward torques from registers to another variable.

fmauch:
I think this PR is totally fine as is.

We could just see how we keep your work in a way that people can reuse it off they want to.

gtatiya:
I tested this PR on URSim and real robot  for UR5 Software 3.7.0, and it works perfectly!! I also tested the master branch since its merged and it also works perfectly!!

image

@fmauch @gavanderhoorn Thank you so much for the help!!

@tonkei0361 Thanks for updating the PR.

gavanderhoorn:
I would recommend we document the fact that the `effort` field contains the current, not the actual torque.

@fmauch: what / where would be the best place/way to do this?

fmauch:
@gavanderhoorn hmm.... 

The repository's README would probably be the most prominent place. But there's no section it would really fit in.

The `ur_robot_driver`'s README would make sense, but I think almost nobody reads it and it contains almost only quite technical documentation.

PR #166, opened 2020-05-01
Use the package type and not the header type as template parameter for communication

fmauch:
When designing this driver we wanted to have all communication objects inherit
from one common `Package` class.

As we want to serve two different protocols (RTDE and Primary/Secondary), we
had this Package class templated with a header type which is different in the
two protocols. With this design decision we could have one common communication
structure (Streams, Pipelines, Producers, Parsers, Consumers) without rewriting
code.

As the thing distinguishing the different protocols was the Header, we
decided to template all the communication objects using HeaderT.

However, as I recently realized, this destroys the possibility to easily create
consumers using the visitor pattern as being done in the `ur_modern_driver`.
With this, there would have to be one root consumer providing abstract methods
for all packages available (over all interfaces).

By templating the communication layer with the type of the actual package
(In terms of RTDEPackage or PrimaryPackage) we can establish a visitor pattern
at protocol level.

@t-schnell it would be nice if you could find some time having a look at this, whether that also makes sense in your opinion.


fmauch:
Note to me: This has to be further tested before merging.

fmauch:
> Note to me: This has to be further tested before merging.

I checked and I seem to not have broken anything :-)

t-schnell:
Looking over it I find no inconsistencies.

The actual primary client/consumer structure we talked about will be implemented in a separate pull request I assume.

PR #167, opened 2020-05-01
Added case label statement

fmauch:
This statement was missing in the clang-format specification. At least with newer clang-format versions this is not being checked properly, otherwise.


fmauch:
Unfortunately, this seems to be available from clang-format-9 and above.

fmauch:
As `clang-format-9` is actually available in ubuntu 18.04 I changed the expected clang-format version, so we can actually merge this :-)

Issue #163, opened 2020-04-23
Forward script execution failures back to the user


fmauch:
The robot's primary interface actually reports errors in executed script code via a `ROBOT_MESSAGE` of type `ROBOT_MESSAGE_TYPE_RUNTIME_EXCEPTION`.

We can catch those and at least print a warning. This way false script code sent to the script interface will not just silently dissolve in the nirvana.

However, from my understanding we cannot know at this point whether this error message actually is triggered by script sent through our driver, which is why I would only print a warning.

gavanderhoorn:
Is the primary interface intended to be used for submitting user scripts?

It was my understanding that it's mostly used by Polyscope.

fmauch:
Well, it is what we use for script commands anyway. (Technically, we use the secondary interface)

The guys at UR suggested to take a look into that solution, so I wanted to write it down here the least.

gavanderhoorn:
O wait. I'm confusing two things. `/ignore`.

urfeex:
Closing this, as this will be tracked in https://github.com/UniversalRobots/Universal_Robots_Client_Library/issues/425

Issue #131, opened 2020-03-04
Publishing script commands doesn't yeild anything


Karthik-Ranga:
# Summary
The issue could be related to #79  #77 
Not able to send any script command through rostopic 

Nothing happens after sending data through script_command

`rostopic pub /ur_hardware_interface/script_command std_gs/String""`

# Versions
 - ROS Driver version: ROS Melodic
 - UR+ product(s) installed: Robotiq Grippers
 - URCaps Software version(s): External Control 1.0.1

# Impact
Unable to send the script files

# Issue details
Sending below

`rostopic pub /ur_hardware_interface/script_command std_msgs/String "data: 'set_digital_out(3,True)'" `

yields nothing. The external control program doesn't stop as said in #79 

## Project status at point of discovered
Found the problem when trying to send script commands to Robotiq gripper.

## Steps to Reproduce
After publishing the script_command,

Checked through
`rostopic echo ur_hardware_interface/script_command`

Received the information
`data: "set_digital_out(2,True)"`

## Expected Behavior
Expected the script file to execute the information

## Actual Behavior
The external control program did not stop and nothing happend

gavanderhoorn:
> The issue is related to #79 #77

Karthik-Ranga:
Agreed. it was also reported in #79 that the program was stopped once the script was sent. But in my case, nothing happened and no one has reported anything related to that so far. I guess it's 2 different problems but common root cause.

Karthik-Ranga:
Just FYI
Not sure if this would help. I got the below info when I checked it. 
```
rosnode info /ur_hardware_interface/script_command
--------------------------------------------------------------------------------
Node [/ur_hardware_interface/script_command]
Publications: None

Subscriptions: None

Services: None

cannot contact [/ur_hardware_interface/script_command]: unknown node
```
When I checked
`rosnode info /ur_hardware_interface`

under subscriptions I found,

```
Subscriptions: 
 * /ur_hardware_interface/script_command [unknown type]
```

fmauch:
`rosnode info /ur_hardware_interface/script_command`

The node is `/ur_hardware_interface` which is why a `rosnode info` on the actual topic doesn't yield anything. This is fine.

One important information is missing: Which robot model and polyscope version are you using? If you are on an e-Series robot you will have to put the robot into remote_control mode in order to make it accept remote script commands.

Karthik-Ranga:
[...]

Issue #118, opened 2020-02-25
Script command by action API


khssnv:
# Enhancement Proposal
Implement an action server API to launch and track arbitrary URScript execution.

# Use case
- track script execution status
- return an error on a script command illegal (if there there is a checker available)
- cancel URScript execution

# Notes
This could be implemented using dirver API already exists (dashboard interface + script command topic).

What do you think about it? Does it make sense for anyone but me?

gavanderhoorn:
This would be nice (in fact I'm implementing this for another robot brand right now).

However, afaik it's impossible to track execution status for a particular program with the current interfaces we have. It's possible to see *some program* is executing, but that would also say "executing" when you're using the UI (as all UI actions are essentially URScriptlets).

Does the dashboard server allow introspection to the level required for this?

fmauch:
> Does the dashboard server allow introspection to the level required for this?

Not that I know. As you said, tracking script code isn't available in any of the interfaces.

khssnv:
Ok, to track what program is actually running and handle action interrupted by another source of command there should be a feedback. There are script TCP/IP communication functions (`socket_open`, `socket_send_byte`, `socket_close`), solution could have a protocol on top of it like it mentioned in #40. Is it a right approach?

fmauch:
It is an approach that helps you in some situations. It gets problematic if your script code doesn't run correctly so you never reach that part. Or if your script code is syntactically wrong.

From a driver's perspective we try to keep things as close to the robot's actual interfaces, as possible. However, adding this as a separate wrapper node might be a valid option.

If you are willing to give it a shot, feel free to open up a PR then we can discuss about it in detail...

khssnv:
> This would be nice (in fact I'm implementing this for another robot brand right now).
@gavanderhoorn, could you share it? Could be handy to estimate how much time it will take for me.

gavanderhoorn:
No, I'm afraid I can't.

urfeex:
This will most likely not be implemented for ROS 1 anymore. However, we still want that feature, but the functionality will be implemented in UniversalRobots/Universal_Robots_Client_Library#425.

Issue #77, opened 2020-01-13
Sending URScript commands for gripper have no effect 


ingmar12345:
# Summary
I am trying to control the Robotiq 2f-85 gripper on the UR5 by sending URScript commands.

I am publishing the commands to the `/ur_hardware_interface/script_command` topic

Example:

`rostopic pub -1 /ur_hardware_interface/script_command std_msgs/String "rq_open()"`

`rq_open()` and `rq_close()` are used to open and close the gripper. The gripper is activated before executing these commands.

The script commands work fine when executing them as separate Script Code nodes on the PolyScope tablet.

Why won't the commands work when publishing them to the `/ur_hardware_interface/script_command` topic?

The ROS driver is running on both the robot side and PC side.

I am running ROS Kinetic on Ubuntu 16.04 Xenial natively on a laptop PC.

# Versions
 - ROS Driver version: 0.0.3
 - Affected Robot Software Version(s): 3.9.1
 - Affected Robot Hardware Version(s): CB3.0 (will upgrade soon to CB3.1)
 - Robot Serial Number: 2015351381
 - UR+ product(s) installed: n/a
 - URCaps Software version(s): Robotiq_grippers 1.2.1, External Control 1.0.1

# Impact
The coupling between the UR5 and the Robotiq gripper will be upgraded soon to a (wireless) interface through the wrist socket, which results in that the gripper cannot be controlled through the official Robotiq driver anymore.

## Use Case and Setup
The objective is to control the Robotiq 2F-85 gripper without the need of the Robotiq driver but in a simple way. That is why I want to use the URscript commands.

## Steps to Reproduce
Make sure the Robotiq 2F-85 Gripper is installed to the UR5 robot, in case of a USB cable, plug it in the controller.
- Run ur_robot_driver on both the the robot side (external control) and ROS side (pc)
- Install the robotiq_grippers URcap, make sure the led turns blue on the Robotiq gripper
- Publish gripper commands (strings) in the following order to the topic called `/ur_hardware_interface/script_command`
`rq_reset()` -> `rq_activate()` -> `rq_open()` -> `rq_close()`

## Expected Behavior
I expected the gripper to perform opening and closing operations.

## Actual Behavior
The gripper didn't move at all.

## Workaround Suggestion
Using the Robotiq driver (http://wiki.ros.org/robotiq), please note that this driver is not compatible with the wireless wrist socket coupling interface for the Robotiq gripper and the UR5.

[...]

PR #171, opened 2020-05-04
Improve primary interface

fmauch:
This branch will
 - Add a full primary client object structure to the client
 - Read more data from the primary interface
 - Give users output when script code sent to the primary interface gets started and finishes as requested in #163, #118, #19, #40 which will hopefully help with #131 #77
 - Show users when they send malformed script code to the robot
 - Inform users when they are trying to send script code to an e-Series robot which is not in remote control mode
 - Output any error messages coming up on the robot

I am not sure yet, to which extent all of those things will be tackled in this MR, but it will definitely be a start from which such features could easily be implemented.

### DIFF
--- a/ur_robot_driver/CMakeLists.txt
+++ b/ur_robot_driver/CMakeLists.txt
@@ -83,11 +83,7 @@ include_directories(
 add_library(ur_robot_driver
     src/comm/tcp_socket.cpp
     src/comm/server.cpp
-    #src/ros/service_stopper.cpp
-    #src/ur/commander.cpp
-    #src/ur/master_board.cpp
-    #src/ur/messages.cpp
-    #src/ur/robot_mode.cpp
+    src/primary/primary_client.cpp
     src/primary/primary_package.cpp
     src/primary/robot_message.cpp
     src/primary/robot_state.cpp
--- a/ur_robot_driver/CMakeLists.txt
+++ b/ur_robot_driver/CMakeLists.txt
@@ -92,6 +92,8 @@ add_library(ur_robot_driver
     src/primary/robot_message.cpp
     src/primary/robot_state.cpp
     src/primary/robot_message/version_message.cpp
+    src/primary/robot_message/text_message.cpp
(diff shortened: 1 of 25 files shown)


gavanderhoorn:
High-level comment: one of the pain points with `ur_driver` and `ur_modern_driver` was the reliance on parsing the binary protocol used on `tcp://:30002` et al. This was a pain point, as the binary protocol changes/changed between updates.

The approach employed here in `ur_robot_driver` was to use RTDE to avoid doing this and thus not have to play catch-up to changes in the protocol as with RTDE the consumer is in control of which fields are part of the messages.

This PR seems to introduce more binary parsing: have measures been put in place to avoid problems with these parsers if/when updates are released, or are the packets parsed expected to "never" be updated?

fmauch:
> High-level comment: one of the pain points with `ur_driver` and `ur_modern_driver` was the reliance on parsing the binary protocol used on `tcp://:30002` et al. This was a pain point, as the binary protocol changes/changed between updates.

Yes, I am very well aware of that, but unfortunately there are features only available in those interfaces. And at least some of them will not be moved to other interface, as for example the controller is only sending script execution feedback on the particular socket connection that has sent the script. The benefit from getting this feature currently outperforms those concerns from my point of view, as many users seem to struggle with sending script code.

> The approach employed here in `ur_robot_driver` was to use RTDE to avoid doing this and thus not have to play catch-up to changes in the protocol as with RTDE the consumer is in control of which fields are part of the messages.
> 
> This PR seems to introduce more binary parsing: have measures been put in place to avoid problems with these parsers if/when updates are released, or are the packets parsed expected to "never" be updated?

This is currently only a draft implementation and no, there's no such thing currently. But it is definitely on the feature list. I plan to do a similar approach as done in the `ur_modern_driver`. In combination with integrated testing at UR (which we currently try to push) we hope to keep the pain at developer's side, at least :-)

fmauch:
Closing in favor of https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/26

Issue #176, opened 2020-05-11
Uninitialized variable in robot_state_helper leads to invalid_argument


RobertWilbrandt:
# Summary

Using the ```set_mode``` action of ```robot_state_helper``` directly after driver startup (e.g. in the same launch file can lead to a ```robot_state_helper``` crash with the following error message:

```console
terminate called after throwing an instance of 'std::invalid_argument'
  what():  Unknown robot mode: -128
```

This is most likely caused by an uninitialized variable.

I will probably create a PR to fix this, but only when i am able to test with a robot again.

# Versions

(I cannot really test this on master, as for some reason nobody gave me a robot arm to take home... but from looking at the code the error seems to still exist)

 - ROS Driver version: version 0.0.3, commit a926789adc0cefc616beb0fda15793cdcd65831b
 - Affected Robot Software Version(s):
 - Affected Robot Hardware Version(s): CB3 UR5
 - Robot Serial Number:
 - UR+ product(s) installed: -
 - URCaps Software version(s): -

# Impact

Not tragic, as this only happens when using a robot startup script (i.e. a script that handles robot startup from the launch file) and a workaround is straightforward.

# Issue details

This is probably caused by the uninitialized variables ```robot_mode_``` and ```safety_mode_``` in [robot_state_helper.cpp](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/src/ros/robot_state_helper.cpp). When the ```set_mode``` action is called with a target state of ```RUNNING``` before a ```/robot_mode``` or ```/safety_mode``` message is received, ```doTransition()``` is called, which goes to default for bot ```robot_mode_``` and ```safety_mode_``` and tries to call ```robotModeString()```, which throws the exception.

## Project status at point of discovered

Unsurprisingly, the issue appeared nondeterministically, but with about 50% likelihood.

## Steps to Reproduce

Start both the ur driver and a startup script in the same launch file. Make the startup script call the ```set_mode``` action immediately.

## Workaround Suggestion

Wait for messages on the ```/robot_mode``` and ```/safety_mode``` topics before calling the action.

gavanderhoorn:
> I cannot really test this on master, as for some reason nobody gave me a robot arm to take home.

`ursim`?

fmauch:
[...]

PR #177, opened 2020-05-11
Initialize member variables

fmauch:
It can happen that the action gets triggered before the mode callback got triggered which will lead to a crash of the helper. (See #176)

While this changes stops the helper from crashing when this happens, it might
not be the best idea to do so as the question remains, what we should do
if we haven't even received a current status from the robot.

With the changes introduced inside this commit, the helper would trigger the
respective state changes, which might lead to wrong requests if we aren't entirely
sure what to do.

One solution would be to reject goals as long as no status was received,
but that would break such scenarios where you want to activate the robot automatically
during startup.

Another idea would be to delay actually starting the action server until we
received both, robot mode and safety mode. But I am not entirely sure whether
this will scale well.

### DIFF
--- a/ur_robot_driver/include/ur_robot_driver/ros/robot_state_helper.h
+++ b/ur_robot_driver/include/ur_robot_driver/ros/robot_state_helper.h
@@ -59,6 +59,7 @@ public:
   virtual ~RobotStateHelper() = default;
 
 private:
+  ros::NodeHandle nh_;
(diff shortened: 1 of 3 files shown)


RobertWilbrandt:
I also used a delayed action server start as a workaround, this worked quite nicely. This would also be my intuitive understanding of correct behavior, as it is obvious that the action call cannot do any work without getting state information (which probably means the robot is still starting up) and clients using the interface would use ```wait_for_server()``` anyway.

PR #179, opened 2020-05-12
Improve documentation

### LINKED ISSUE #38: Document / change blocking behavior of read() function
In #34 it was raised that the robot controller dictates the loop rate by having implemented the `read()` function blocking.

This was in fact a design choice, but I realize, that this is an issue when using this driver in a combined_hw interface.

In a first step we should at least document this. A second step could be to take a step back and think about the advantages, disadvantages and alternatives to the implemented approach with keeping both, close-real-time behavior and combined_hw support in mind.

### LINKED ISSUE #51: Add documentation about different modes
We currently have two modes: "Normal" mode which requires the URCap being installed on the robot and the "External Control" program node running inside a program. The "headless" mode on the other side sends the primary URScript for execution to the robot directly.

An orthogonal mode to those two is remote-control mode. e-Series robots have to be put into that mode to make certain features such as sending script code and most of the dashboard commands work.

All this is not documented all too well currently.

### LINKED ISSUE #140: Why can't the driver use the extracted calibration info on startup?
# Summary

Currently, there is a two-step process for running an arm with calibration correction. First, you run `roslaunch ur_calibration calibration_correction.launch robot_ip:=<robot_ip> output_filename:=/home/user/calibration.yaml` to extract the calibration information. 

Then, to use an arm with calibration correction, you'd run `roslaunch ur_robot_launch ur10_bringup.launch robot_ip:=<robot_ip> kinematics_config:=/home/user/calibration.yaml`

From what I can tell, the driver then checks that the `calibration.yaml` file (generated in step 1) is for the correct arm by extracting the calibration information from the arm again, creating a checksum by concatenating these values, and comparing that checksum with the checksum in the `calibration.yaml` file. 

My question is, if the checksum that the driver creates to validate the calibration file is generated from freshly extracted calibration information, why can't this information be used to automatically apply calibration correction to the arm when it starts? Why the need for the two step process?

fmauch:
This MR aims at collecting a couple of documentation improvements.

Closes #140 

Closes #38  

Closes #51

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,3 +261,8 @@ Additionally: If the calibration doesn't match the expected one, there's a reaso
 should better be explicitly handled by a human. Having to run the calibration
 extraction/transformation as a separate step makes this possible and doesn't hide this step from the
 end user.
+
+### Can this driver be used inside a combined hardware interface?
+Yes, this is possible. However, if used inside a [combined HW
+interface](http://wiki.ros.org/combined_robot_hw) we recommend to enable [non-blocking read
+functinality](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/doc/ROS_INTERFACE.md#non_blocking_read-default-false).
--- a/README.md
+++ b/README.md
@@ -251,13 +251,13 @@ the computations requiring the graphics card.
 This is mainly because parameters are loaded onto the parameter server before any nodes are started.
 
 The `robot_description` concept inside ROS is not designed to be changed while a system is running.
-Consumers of the urdf/`robot_description` will not update the model they have been loading
-initially. It's not the driver that needs/benefits from this calibrated urdf, it's the rest of the
-ROS application.
+Consumers of the urdf/`robot_description` (in terms of other ROS nodes) will not update the model
+they have been loading initially. While technically the description could be altered during runtime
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -246,3 +246,18 @@ This is a known issue and unfortunately we don't have a solution for this. The N
...
+Consumers of the urdf/`robot_description` will not update the model they have been loading
+initially. It's not the driver that needs/benefits from this calibrated urdf, it's the rest of the
+ROS application.
+
+Additionally: it's good to have this sort of things done in an off-line fashion, as that makes it
+predictable. If/when, for whatever reason, the calibration data changes, it's almost always better
+to have the files updated because of a conscious decision to do that, instead of automatically,
+invisible.  Having to run the calibration extraction/transformation as a separate step makes this
+possible. If the calibration doesn't match the expected one, there's a reason for that and it should
+better be explicitly handled by a human.
```
gavanderhoorn, line 263: I know I wrote it, but I suggest to reword this such that it fits better in an FAQ.

Right now this reads like how I would explain something to a colleague (almost), which is too casual almost.

fmauch: I thought I already did a proper rewording, at least for the non-additional part :-D

gavanderhoorn: hmm :thinking:  ;)

fmauch: I did a second try :-)

PR #193, opened 2020-06-12
Add actual documentation link into calibration checker output

fmauch:
The output was generated when we didn't have the final repository available.
However, updating the output got lost over time...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,3 +29,16 @@ jobs:
       - uses: actions/checkout@v1
       - uses: 'ros-industrial/industrial_ci@master'
         env: ${{matrix.env}}
+
+
+  check_urls:
+    runs-on: ubuntu-latest
+    steps:
+     - uses: actions/checkout@v1
+     - name: check URLs
+       run: |
+         echo "$(pwd)"
+         ls -l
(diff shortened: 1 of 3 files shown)


gavanderhoorn:
Would there be any way for us to track when this link breaks (ie: because you move sections around, move them to other files or rename the section)?

fmauch:
I guess this would require a ci job that extracts the link from source code and checks it...

gavanderhoorn:
MoveIt uses [html-proofer](https://github.com/gjtorikian/html-proofer) for this: [here](https://github.com/ros-planning/moveit_tutorials/blob/48f340e59ba1ced913ec4b2c8d5034c19ad2e0ae/.travis.yml#L50-L52).

Might be something to look into?

fmauch:
I just used a simple bash script to achieve what I want.

Ideally we would let run this over the whole repository, but I found that we currently have a couple of invalid (as in 404) links in our codebase. I'll leave this for a separate PR, though.

fmauch:
I'll merge this and add checking the other links separately.

gavanderhoorn:
@fmauch: would it perhaps be better to add the `check_urls.sh` script to the `.github` directory?

`scripts` contains scripts used by users of the driver. The check url script seems out of place there.

fmauch:
True. I'll keep that in mind when checking the other links, as well.

PR #197, opened 2020-06-16
robot_driver: use pass_all_args to reduce verbosity

gavanderhoorn:
As per subject.

Note: I haven't tested this (as in: started the driver), but I expect it to work as described.

### DIFF
--- a/ur_robot_driver/launch/ur10_bringup.launch
+++ b/ur_robot_driver/launch/ur10_bringup.launch
@@ -13,20 +13,7 @@
   <arg name="limited" default="false" doc="Use the description in limited mode (Every axis rotates from -PI to PI)"/>
   <arg name="headless_mode" default="false" doc="Automatically send URScript to robot to execute. On e-Series this does require the robot to be in 'remote-control' mode. With this, the URCap is not needed on the robot."/>
 
-  <include file="$(find ur_robot_driver)/launch/ur_common.launch">
-    <arg name="debug" value="$(arg debug)"/>
+  <include file="$(find ur_robot_driver)/launch/ur_common.launch" pass_all_args="true">
     <arg name="use_tool_communication" value="false"/>
-    <arg name="controller_config_file" value="$(arg controller_config_file)"/>
-    <arg name="robot_description_file" value="$(arg robot_description_file)"/>
-    <arg name="robot_ip" value="$(arg robot_ip)"/>
-    <arg name="reverse_port" value="$(arg reverse_port)"/>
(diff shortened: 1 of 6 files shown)


ur_robot_driver/launch/ur10_bringup.launch
```diff
@@ -13,20 +13,7 @@
   <arg name="limited" default="false" doc="Use the description in limited mode (Every axis rotates from -PI to PI)"/>
   <arg name="headless_mode" default="false" doc="Automatically send URScript to robot to execute. On e-Series this does require the robot to be in 'remote-control' mode. With this, the URCap is not needed on the robot."/>
 
-  <include file="$(find ur_robot_driver)/launch/ur_common.launch">
-    <arg name="debug" value="$(arg debug)"/>
+  <include file="$(find ur_robot_driver)/launch/ur_common.launch" pass_all_args="true">
     <arg name="use_tool_communication" value="false"/>
```
gavanderhoorn, line 17: It could be nice to move `use_tool_communication` to the list of `args` at the top of the file and instead of providing a `default`, just set the `value` (like it's done here).

As it is right now though it does clearly communicate the fact that for these robots (ie: non-e-Series), `use_tool_communication` cannot be set to anything but `false`. Perhaps adding a comment explaining *why* this is would be good, to avoid users getting confused when they try to edit the `.launch` file and force it to `true`.

fmauch:
> As per subject.
> 
> Note: I haven't tested this (as in: started the driver), but I expect it to work as described.

I did not know, this was possible! That should simplify maintenance a lot.

gavanderhoorn:
> I did not know, this was possible! That should simplify maintenance a lot.

has been supported since Indigo, so ..

Perhaps you're not paying attention ;)

fmauch:
> > I did not know, this was possible! That should simplify maintenance a lot.
> 
> has been supported since Indigo, so ..
> 
> Perhaps you're not paying attention ;)

Yes, I've just seen that in the docs. I think, I never read the docs for the `include` tag, as it is basically quite simple to use... But I agree that I should have.

ur_robot_driver/launch/ur10_bringup.launch
fmauch: Should we incorporate this here or a separate PR?

gavanderhoorn: I would do it in a separate one (as it's a different subject).

But you're the maintainer, so you could *request changes*.

fmauch: Let's keep this separate, then.

PR #207, opened 2020-06-25
WIP: make variables global

ipa-lth:
Explicitly state globals (then also works when within function).
This change enables to have more than one node in the ur-program.
Closely relates to https://github.com/UniversalRobots/Universal_Robots_ExternalControl_URCap/issues/5 and https://github.com/UniversalRobots/Universal_Robots_ExternalControl_URCap/pull/7

Quoting the scriptmanual_en.pdf, section 1.7 (p. 6):
>Every variable declared inside a program exits at a global scope, except when they
are declared inside a function. In that case the variable are local to that function. Two
qualifiers are available to modify this behaviour. The local qualifier tells the runtime to
treat a variable inside a function, as being truly local, even if a global variable with the
same name exists. The global qualifier forces a variable declared inside a function,
to be globally accessible.

### DIFF
--- a/ur_robot_driver/resources/ros_control.urscript
+++ b/ur_robot_driver/resources/ros_control.urscript
@@ -3,7 +3,7 @@
 global steptime = get_steptime()
 
 textmsg("steptime=", steptime)
-global MULT_jointstate = {{JOINT_STATE_REPLACE}}
+MULT_jointstate = {{JOINT_STATE_REPLACE}}
 
 #Constants
 global SERVO_UNINITIALIZED = -1
--- a/ur_robot_driver/resources/ros_control.urscript
+++ b/ur_robot_driver/resources/ros_control.urscript
@@ -1,20 +1,20 @@
 {{BEGIN_REPLACE}}
 
-steptime = get_steptime()
+global steptime = get_steptime()
 
 textmsg("steptime=", steptime)
-MULT_jointstate = {{JOINT_STATE_REPLACE}}
+global MULT_jointstate = {{JOINT_STATE_REPLACE}}
 
 #Constants
-SERVO_UNINITIALIZED = -1
-SERVO_IDLE = 0
-SERVO_RUNNING = 1
+global SERVO_UNINITIALIZED = -1
+global SERVO_IDLE = 0
+global SERVO_RUNNING = 1
 
-MODE_STOPPED = -2
-MODE_UNINITIALIZED = -1
-MODE_IDLE = 0
-MODE_SERVOJ = 1
-MODE_SPEEDJ = 2
+global MODE_STOPPED = -2
+global MODE_UNINITIALIZED = -1
+global MODE_IDLE = 0
+global MODE_SERVOJ = 1
+global MODE_SPEEDJ = 2
 
 #Global variables are also showed in the Teach pendants variable list
 global cmd_servo_state = SERVO_UNINITIALIZED


fmauch:
This PR would only make sense, of the code would be wrapped into a function, which is not yet decided in https://github.com/UniversalRobots/Universal_Robots_ExternalControl_URCap/issues/5.

ipa-lth:
> 
> 
> This PR would only make sense, of the code would be wrapped into a function, which is not yet decided in [UniversalRobots/Universal_Robots_ExternalControl_URCap#5](https://github.com/UniversalRobots/Universal_Robots_ExternalControl_URCap/issues/5).

Yes, but it does not harm with the current URCAP version.
> 
> Also: Is this necessary, with the implementation of [UniversalRobots/Universal_Robots_ExternalControl_URCap#7](https://github.com/UniversalRobots/Universal_Robots_ExternalControl_URCap/pull/7)?

It is. With the above PR, it won't work.
> 
> **Edit:** Yes, it is. I was assuming that script code like this would work:
> 
> ```python
> def my_fun():
>   my_var = "World"
>   def nested_function():
>     textmsg("Hello" , my_var)
>   end
>   nested_function()
> end
> my_fun()
> ```

I was assuming the same thing, turns out: No.
See the manual, section 1.7:
> A urscript program is declared as a function without parameters:
> ```python
> def myProg():
> end
> ```
> Every variable declared inside a program exits at a global scope, except when they
are declared inside a function. In that case the variable are local to that function. Two
qualifiers are available to modify this behaviour. The _local_ qualifier tells the runtime to
treat a variable inside a function, as being truly local, even if a global variable with the
same name exists. The _global_ qualifier forces a variable declared inside a function,
to be globally accessible.

So, the assumption of the scope of the variables only hold true for the ur-program (function) itself. Otherwise, this changeset would definitely not be necessary. It seems that the "root" function has special properties.

ur_robot_driver/resources/ros_control.urscript
```diff
@@ -1,20 +1,20 @@
 {{BEGIN_REPLACE}}
 
-steptime = get_steptime()
+global steptime = get_steptime()
 
 textmsg("steptime=", steptime)
-MULT_jointstate = {{JOINT_STATE_REPLACE}}
+global MULT_jointstate = {{JOINT_STATE_REPLACE}}
```
fmauch, line 6: I think, that one can skip the `global` keyword.

ur_robot_driver/resources/ros_control.urscript
```diff
@@ -1,20 +1,20 @@
 {{BEGIN_REPLACE}}
 
-steptime = get_steptime()
+global steptime = get_steptime()
```
fmauch, line 3: We could probably replace all occurrences of `steptime` with `get_steptime()` to prevent this global variable. I am unsure, however, whether this will have impact on the script performance.

ipa-lth: I had the same thought and the same doubt...

ur_robot_driver/resources/ros_control.urscript
ipa-lth: You are right, i removed the statement

fmauch:
Closing this in favor of https://github.com/UniversalRobots/Universal_Robots_ExternalControl_URCap/pull/14

@ipa-lth If you like, go ahead and test it. If you disagree that this solves the problem addressed here, please comment.

PR #110, opened 2020-02-13
F parameterise using local hardware interface param

carebare47:
In it's current form, this driver won't allow multiple arms to exist in the same namespace. This pr fixes that by adding a parameter that allows the hardware_interface/joints parameter to be loaded from the drivers local namespace instead of the global one.

PR #217, opened 2020-07-03
Add prefix to wrench hw interface

carebare47:
Adds a prefix to the wrench hw interface so that multiple arms can have their force torque sensors used at the same time


fmauch:
I think, it would be better to either use the `tf_prefix` or use a parameter for the complete interface name.

ur_robot_driver/src/ros/hardware_interface.cpp
```diff
@@ -89,6 +90,9 @@ bool HardwareInterface::init(ros::NodeHandle& root_nh, ros::NodeHandle& robot_hw
   // to publish correct frame names for frames reported by the robot directly.
   robot_hw_nh.param<std::string>("tf_prefix", tf_prefix_, "");
 
+  // Optional parameter to add a prefix to the 'wrench' hardware interface (if using more than one arm)
+  robot_hw_nh.getParam("wrench_prefix", wrench_prefix);
```
fmauch, line 94: Please use the syntax from the parameter above with a default value if the parameter is optional

carebare47: 99825fbbf40f6080f81289a12a0f9e2760f1ec2c

carebare47:
> I think, it would be better to either use the `tf_prefix` or use a parameter for the complete interface name.

6e21e0d1e44e0ce9b3b30363e42cba187a766be7

fmauch:
Do you think, it would make sense to read it from the root node handle, similar to [joint_names](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/src/ros/hardware_interface.cpp#L281), then we could configure this together with instanciating the controller in https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/config/ur10_controllers.yaml#L21

In my opinion that would be preferable over introducing yet another launch file parameter.

fmauch:
Also, once everything is in place, we should make sure that documentation is everywhere at the right places, e.g. [here](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/doc/ROS_INTERFACE.md)

carebare47:
> Do you think, it would make sense to read it from the root node handle, similar to [joint_names](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/src/ros/hardware_interface.cpp#L281), then we could configure this together with instanciating the controller in https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/config/ur10_controllers.yaml#L21
> 
> In my opinion that would be preferable over introducing yet another launch file parameter.

I think that would cause us a similar problem to the one we have here: https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/110

gavanderhoorn:
This doesn't appear to be setting a prefix any more, but it configures the actual `frame_id` value.

Suggestion: use `wrench_frame_id` as the name everywhere (ie: parameter name, local member variable, etc).

carebare47:
> This doesn't appear to be setting a prefix any more, but it configures the actual `frame_id` value.
> 
> Suggestion: use `wrench_frame_id` as the name everywhere (ie: parameter name, local member variable, etc).

That's nicer, thanks

carebare47:
> Also, once everything is in place, we should make sure that documentation is everywhere at the right places, e.g. [here](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/doc/ROS_INTERFACE.md)

91d31f5dc2e9c750ffbd566d36e2660b872c6c4b

Issue #151, opened 2020-03-30
Consider lowering action server state reporting frequency


gavanderhoorn:
@fmauch wrote in https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/issues/138#issuecomment-605828287:
> I think the `state_publish_rate` is the same as the control loop rate only for legacy reasons as this was the same in the `ur_modern_driver` for CBx series and every 125 Hz was replaced by 500Hz and later by `*loop_hz` without putting too much thought into that...

We might want to consider lowering the `state_publish_rate` of the various actionlib based controllers.

As these are typically long(er) running actions with more discrete state changes, publishing their state (ie: those of the actions, not of the robot) at 500 Hz seems wasteful (as the state has probably not changed sufficiently to be of interest).

gavanderhoorn:
I can quickly fix this, but a bit of discussion about a rate would be good.

The default is 25 Hz I believe.

Would 50 Hz be acceptable here?

It would still be something users could change, would they so desire.

fmauch:
According to [the docs](http://wiki.ros.org/joint_trajectory_controller#Parameters) the default value is 50 Hz. I think, it would be fine to stick with that.

gavanderhoorn:
> According to [the docs](http://wiki.ros.org/joint_trajectory_controller#Parameters) the default value is 50 Hz. I think, it would be fine to stick with that.

It is: [ros-controls/ros_controllers/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h](https://github.com/ros-controls/ros_controllers/blob/1fd8d25707e289700743925c93644b4295b1c486/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L218).

I'll set it to that value.

gavanderhoorn:
See #221.

fmauch:
Closed via #221

PR #221, opened 2020-07-07
driver: use default rates for JTC goal monitor rates.

gavanderhoorn:
As per subject.

See the commit comment for some rationale.

Note: the `state_publish_rate` has been left to be equal to the `loop_hz`, as this controls how often the internal (position/velocity/effort) controller of the JTC publishes its state (ie: tracking error, target position, etc). That is something which *should* be published at the same rate the controller is run to actually be useful.

### DIFF
--- a/ur_robot_driver/config/ur10_controllers.yaml
+++ b/ur_robot_driver/config/ur10_controllers.yaml
@@ -43,7 +43,7 @@ scaled_pos_traj_controller:
       wrist_3_joint: {trajectory: 0.2, goal: 0.1}
    stop_trajectory_duration: 0.5
    state_publish_rate: *loop_hz
-   action_monitor_rate: 10
+   action_monitor_rate: 20
 
 pos_traj_controller:
    type: position_controllers/JointTrajectoryController
@@ -59,7 +59,7 @@ pos_traj_controller:
       wrist_3_joint: {trajectory: 0.2, goal: 0.1}
    stop_trajectory_duration: 0.5
    state_publish_rate: *loop_hz
-   action_monitor_rate: 10
+   action_monitor_rate: 20
 
 scaled_vel_traj_controller:
    type: velocity_controllers/ScaledJointTrajectoryController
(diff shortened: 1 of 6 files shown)


gavanderhoorn:
This is for #151.

And for `wrid20`.

fmauch:
I am  a bit confused. I though you were talking about `state_publish_rate` all the time. The `action_monitor_rate` never was at `loop_hz` but at 10 Hz as far as I can see.

If I read the diff correct, this PR basically increases the `action_monitor_rate` from 10 to 20 Hz. Do I get something wrong here?

gavanderhoorn:
No. That's what I did.

Please see the commit comment.

It should clarify things.

It doesn't make sense to lower `state_publish_rate` actually.

It does make sense to increase `action_monitor_rate`.

So that's what I did.

fmauch:
> No. That's what I did.
> 
> Please see the commit comment.

I did, that's part of my confusion

> These were set to run at the 'loop rate' of the controller, which would be 125 Hz and 500 Hz for the CB3 and e-Series respectively.
>
> As these rates are mostly concerned with publishing the state of the goal, not the internal state of the controller, and the goal state is mostly concerned with whether a trajectory has completed execution, was cancelled or preempted, running this at 500 hz was deemed a bit too much.

That's the part I don't understand. Why were they running at `loop_hz` before making those changes?

> It doesn't make sense to lower state_publish_rate actually.

I agree :-)

> It does make sense to increase action_monitor_rate.

I agree :-)

> So that's what I did.

I'm fine with merging the result.

gavanderhoorn:
> Why were they running at `loop_hz` before making those changes?

Yes, that is confusing. I wrote that comment when I still had a change in there for `state_publish_rate`.

I'll update the commit comment.

fmauch:
@gavanderhoorn Thank you, this is more understandable :-)

gavanderhoorn:
"Fixed"

branch name is still a bit weird, but I'll leave it like this.

PR #227, opened 2020-07-14
Use Robot_hw_nh node handle for joints.

CollinAvidano:
Modified hardware interface to look for joints parameter under the robot_hw node handle. By doing this, removing the hardware_interface prefixed namespace and just having joints as the parameter name, I have managed to get this working well with both a combined robot hardware interface with 2 arms under the same controller manager and with a single arm as is usual.

Similar to pr #110 but does not require the addition of another param. Does modify the hardware_interface param in the controller yamls to be ur_hardware_interface. But after testing and looking at source it seems only the joints parameter references that name.

### DIFF
--- a/ur_robot_driver/config/ur10_controllers.yaml
+++ b/ur_robot_driver/config/ur10_controllers.yaml
@@ -3,7 +3,7 @@ hardware_control_loop:
    loop_hz: &loop_hz 125
 
 # Settings for ros_control hardware interface
-hardware_interface:
+ur_hardware_interface:
    joints: &robot_joints
      - shoulder_pan_joint
      - shoulder_lift_joint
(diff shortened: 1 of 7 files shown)


fmauch:
Thanks for the contribution. On a first glance I would say, this does make sense. I think, this is the proper solution for #110. @carebare47 would you agree with this solution?

fmauch:
waiting for an answer from @carebare47 before merging

CollinAvidano:
> Thanks for the contribution. On a first glance I would say, this does make sense. I think, this is the proper solution for #110. @carebare47 would you agree with this solution?

Np! And thank you to you all for writing the rest of this! Figured the least I could to was contribute a parameter change I use.

fmauch:
@carebare47 I am going to merge this by the end of the week and close #110 if no further comments are coming in.

CollinAvidano:
Just wanted to ask is this going to be merged?

fmauch:
Thanks @CollinAvidano for your contribution!

PR #228, opened 2020-07-22
Changed network in Industrial CI pipeline

ancker1:
This Pull-Request contains changes to the network used in Industrial CI pipeline and the static IPs used in the tests.

This Pull-Request moves away from using the default Docker network in the range 172.17.0.0/16, since a specific IP cannot be chosen/guaranteed within this range. Instead, a network is now created with range 192.168.0.0/16, which allows for assigning specific IPs to the containers.


gavanderhoorn:
I'd be curious to know the context which led @ancker1 to propose these changes.

fmauch:
@ancker1 is working on integrating the testing pipeline into regular testing at UR directly :-). We agreed to keep the internal testing as close as possible to the one we use in this repo for maintenance reasons.

PR #245, opened 2020-08-13
Updated externalcontrol to v1.0.3

### LINKED ISSUE #237: UR10e  - huge lag when programs loaded/played from the Ubuntu PC
I want to load and run programs that I have on the UR. The ROS PC has a perfect connection to the UR. When the load command is sent the programs gets loaded. When we run it. The UR gives a popup on the polyscope that says "starting program". Before we installed the external control URcap and the gripper (schunck EGH-80) URcap the programs would run immediately after pressing play. Since these URcaps the delay is huge. The ROS driver times out after one second. So we increased the timeout period. But that is no solution as simple motion programs are taking as much as a few minutes to load and play. On the DoF forum I saw two similar queries with ROS and Gripper URcaps. But they arent answered yet. Any body here has an idea how I can get the execution in a respectable time. Thanks

fmauch:
See the urcap's release notes [here](https://github.com/UniversalRobots/Universal_Robots_ExternalControl_URCap/releases/tag/v1.0.3).

I'll leave this unmerged for a couple of days until I have tested it more thoroughly.

Closes #237


fmauch:
I found a bug in v1.0.3 that got fixed in v1.0.4, so updating this MR. With that version I feel very confident, so I will merge this eod.

PR #259, opened 2020-09-08
testing: Use a valid IP address with no robot endpoint

fmauch:
Using an IP address from a valid ip range makes the test fail earlier.


fmauch:
Closing this in favor of #266 This test will be running in the library.

PR #266, opened 2020-09-14
Separation

fmauch:
The core library for communicating with Universal Robots manipulators has been [moved to a separate repository](https://github.com/UniversalRobots/Universal_Robots_Client_Library).

This PR removes it from this repository to keep maintenance of library code at one place.

This will be merged as soon as binary packages are available for ROS kinetic, melodic and noetic.

See https://github.com/ros/rosdistro/pull/26529 for the melodic release.

PR #268, opened 2020-09-17
Allow to set the reverse_ip manually, useful to run in a container

fran6co:
When running inside a container (e.g.: Docker) the ip will be an internal one but the port will be binded to the host interface instead. Allowing to set it manually fixes this issue

### DIFF
--- a/ur_robot_driver/include/ur_robot_driver/ur/ur_driver.h
+++ b/ur_robot_driver/include/ur_robot_driver/ur/ur_driver.h
@@ -70,7 +70,7 @@ public:
            const std::string& input_recipe_file, std::function<void(bool)> handle_program_state, bool headless_mode,
            std::unique_ptr<ToolCommSetup> tool_comm_setup, const std::string& calibration_checksum = "",
            const uint32_t reverse_port = 50001, const uint32_t script_sender_port = 50002, int servoj_gain = 2000,
-           double servoj_lookahead_time = 0.03, bool non_blocking_read = false);
+           double servoj_lookahead_time = 0.03, bool non_blocking_read = false, const std::string& reverse_ip = "");
   /*!
    * \brief Constructs a new UrDriver object.
    *
@@ -94,10 +94,10 @@ public:
            const std::string& input_recipe_file, std::function<void(bool)> handle_program_state, bool headless_mode,
            const std::string& calibration_checksum = "", const uint32_t reverse_port = 50001,
            const uint32_t script_sender_port = 50002, int servoj_gain = 2000, double servoj_lookahead_time = 0.03,
-           bool non_blocking_read = false)
+           bool non_blocking_read = false, const std::string& reverse_ip = "")
     : UrDriver(robot_ip, script_file, output_recipe_file, input_recipe_file, handle_program_state, headless_mode,
                std::unique_ptr<ToolCommSetup>{}, calibration_checksum, reverse_port, script_sender_port, servoj_gain,
-               servoj_lookahead_time, non_blocking_read)
+               servoj_lookahead_time, non_blocking_read, reverse_ip)
   {
   }
 
--- a/ur_robot_driver/launch/ur_common.launch
+++ b/ur_robot_driver/launch/ur_common.launch
@@ -4,6 +4,7 @@
   <arg name="use_tool_communication" doc="On e-Series robots tool communication can be enabled with this argument"/>
   <arg name="controller_config_file" doc="Config file used for defining the ROS-Control controllers."/>
   <arg name="robot_ip" doc="IP address by which the robot can be reached."/>
+  <arg name="reverse_ip" default="" doc="IP of the driver, if set to empty it will detect it automatically."/>
   <arg name="reverse_port" default="50001" doc="Port that will be opened by the driver to allow direct communication between the driver and the robot controller."/>
   <arg name="script_sender_port" default="50002" doc="The driver will offer an interface to receive the program's URScript on this port. If the robot cannot connect to this port, `External Control` will stop immediately."/>
   <arg name="kinematics_config" doc="Kinematics config file used for calibration correction. This will be used to verify the robot's calibration is matching the robot_description."/>
@@ -37,6 +38,7 @@
     <arg name="controller_config_file" value="$(arg controller_config_file)"/>
     <arg name="robot_ip" value="$(arg robot_ip)"/>
     <arg name="reverse_port" value="$(arg reverse_port)"/>
+    <arg name="reverse_ip" value="$(arg reverse_ip)"/>
     <arg name="script_sender_port" value="$(arg script_sender_port)"/>
     <arg name="kinematics_config" value="$(arg kinematics_config)"/>
     <arg name="tf_prefix" value="$(arg tf_prefix)"/>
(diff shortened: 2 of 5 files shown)


fmauch:
Shouldn't it still be the same IP address as the one that the RTDEClient uses? This is where the IP address is currently coming from. Could you elaborate your network setup, please?

fran6co:
I have a dockerized driver in a bridge network with it’s own private ips
but the reverse_port is binded to the host interface that has a different
ip. But the ur driver only sees the dockerized ip and not the host ip.

fran6co:
Just for the sake of having a more clear picture:

ur_driver - docker bridge network (ip 192.168.1.21, reverse_port: 50003)
host machine - ethernet interface (ip 192.168.7.66)
ur sim - virtuallbox running in a host-only interface (ip 192.168.56.101, sees the host with ip 192.168.56.1)
port 50003 of the dockerized ur_driver is binded to host machine 0.0.0.0:50003 so any interface can connect to it

The problem is that the dockerized ur_driver is not aware of the binding, and resolves the ip to 192.168.1.21 and sends it to the robot. That ip is not visible to anybody else but the host running that dockerized instance.
ur_sim tries to connect to 192.168.1.21:50003 and fails when it should be connecting to 192.168.56.1:50003 instead.
Being able to specify the ip manually it will fix this

gavanderhoorn:
Manually overriding the IP would work here, but I'm wondering: if you have a bridge network for your containers, why still bind the port from the container to the host? I have a similar setup and "all" that's needed for hosts on the regular network (so connected to `192.168.7.66` in your case) is to know how to route packets for `192.168.1.21` back to the host running the Docker container.

I do that with a static route which is distributed by the router for that segment.

Or do you still have Docker's NAT between your bridge network and the rest of the network?

fran6co:
Yes, a route would solve it in my case but it would expose the private network where I'm running all the dockerized nodes to the rest of the network. That does go against the idea of dockerizing things this way, the robot should only see the host and not be aware that there is a another network.
And also this would be helpful in NAT setups as well, as you mention, where we can't just route that ip.
Not sure what is the objection to this patch, having alternative ways to do it doesn't invalidate the usefulness of being able to specify the parameter manually.

gavanderhoorn:
> Not sure what is the objection to this patch

There is none, neither was I trying to formulate one.

I'm not sure why you are assuming there are objections.

We're still just discussing the merits afaiu.

fran6co:
Sorry, I thought that this was a straightforward change that wouldn't have required much discussion. But I don't mind clarify the use cases as much as you need.

Regarding NAT, I'm actually considering going with the NAT solution instead of the bridged one for improved isolation. In that case the static route wouldn't help

gavanderhoorn:
Similar PR in `ur_modern_driver`: ros-industrial/ur_modern_driver#301.

fran6co:
Why wouldn't you recommend running it via a NAT? I do like the isolation that it provides of each of the components and the extra routing shouldn't be much of an overhead given that it happens in the same machine

gavanderhoorn:
Because Docker's NAT is notoriously slow (or at least it used to be, last time I checked), and this driver employs a real-time external control interface.

Any device between the driver and the controller will increase jitter, which reduces control performance.

gavanderhoorn:
@fmauch: just making sure you're not waiting on some final comment from me.

fmauch:
> @fmauch: just making sure you're not waiting on some final comment from me.

I am not :-)

I'm still not a huge fan introducing yet another parameter to the `UrDriver` class, but I see the point.

However, before taking this one step further, I think, it makes sense to merge #266 first and then build ontop of that. This currently waits for the next melodic sync to me merged.

fmauch:
I forgot about this one a little bit...

Now that #268 is merged, this PR has to be partially implemented inside [the library](https://github.com/UniversalRobots/Universal_Robots_Client_Library). @fran6co would you still like to contribute this?

fmauch:
I thought about it a bit more and I think it would be better to handle this inside the external_control URCaps.

Inside there we know the IP address of the external PC from the configuration. We could inject this into the program code as a global variable and then use that variable inside the URScript portion. This way, we could avoid adding yet another parameter to the `UrDriver` class and the user would only have to configure the IP address once on the TP which (s)he needs to do anyway in order to receive the UrScript in the first place.

**However**, while writing this, I see that for the headless mode we would need to figure that out differently, as the headless mode does not require having the external_control URCap installed so far.

fran6co:
I don't need this right now because the need for containers got pushed pack, so I'm closing it for now

JS00000:
> I don't need this right now because the need for containers got pushed pack, so I'm closing it for now

I think this PR is necessary because I cannot connect the 

> I thought about it a bit more and I think it would be better to handle this inside the external_control URCaps.
> 
> Inside there we know the IP address of the external PC from the configuration. We could inject this into the program code as a global variable and then use that variable inside the URScript portion. This way, we could avoid adding yet another parameter to the `UrDriver` class and the user would only have to configure the IP address once on the TP which (s)he needs to do anyway in order to receive the UrScript in the first place.
> 
> **However**, while writing this, I see that for the headless mode we would need to figure that out differently, as the headless mode does not require having the external_control URCap installed so far.

The reverse_ip is a problem when I run ur5_bringup.launch in the container of Docker for Windows. In ubuntu I can run docker with command `docker run -it -d --name ur5_container --net=host ros_image` and docker will publish all the container's port to the host. But this command will fail in Docker for Windows because windows does not support `--net=host`. I then run docker with command `docker run -it -d --name ur5_container -p 50001-50002:50001-50002 ros_image`, and fail again. The log in Teach Pendant show `Not able to open socket reverse_socket` to host: 172.17.0.2 at port 50001: Connection timed out.

JS00000:
> I thought about it a bit more and I think it would be better to handle this inside the external_control URCaps.
> 
> Inside there we know the IP address of the external PC from the configuration. We could inject this into the program code as a global variable and then use that variable inside the URScript portion. This way, we could avoid adding yet another parameter to the `UrDriver` class and the user would only have to configure the IP address once on the TP which (s)he needs to do anyway in order to receive the UrScript in the first place.
> 
> **However**, while writing this, I see that for the headless mode we would need to figure that out differently, as the headless mode does not require having the external_control URCap installed so far.

The reverse_ip is a problem when I run ur5_bringup.launch in the container of Docker for Windows. In ubuntu I can run docker with command `docker run -it -d --name ur5_container --net=host ros_image` and docker will publish all the container's port to the host. The ur5_bringup.launch will successfully connect to the robot. But this command will fail in Docker for Windows because windows does not support `--net=host`. I then run docker with command `docker run -it -d --name ur5_container -p 50001-50002:50001-50002 ros_image`, and fail to launch ur5_bringup.launch in it. The log in Teach Pendant show `Not able to open socket reverse_socket to host: 172.17.0.2 at port 50001: Connection timed out.` So how should this problem be solved?

fmauch:
@JS00000 Do you have any opinion to the points I mentioned above? Would you like to follow up this PR inside the library?

JS00000:
> @JS00000 Do you have any opinion to the points I mentioned above? Would you like to follow up this PR inside the library?

I think the solution of this PR is a good idea. I will test it later.

PR #287, opened 2020-11-11
Wait for response on reverse socket read

giusebar:
We are currently running multiple UR arms in the same control loop.
Given the architecture of the driver and the presence of blocking calls in initialisation, the _socket_read_binary_interger()_ function times out (the default timeout is 2 seconds according to documentation).
In order to fix this, I'm proposing to change the default timeout to 0 and let the socket just wait for a response.
This allows us to run multiple URs in the same control loop and it does not affect launching a single arm either.

Let me know your thoughts.


gavanderhoorn:
Is the history supposed to be this 'messy'?

PR #288, opened 2020-11-11
Wait for reverse socket response

giusebar:
We are currently running multiple UR arms in the same control loop.
Given the architecture of the driver and the presence of blocking calls in initialisation, the _socket_read_binary_interger()_ function times out (the default timeout is 2 seconds according to documentation).
In order to fix this, I'm proposing to change the default timeout to 0 and let the socket just wait for a response.
This allows us to run multiple URs in the same control loop and it does not affect launching a single arm either.

Let me know your thoughts.


fmauch:
Yes, we could certainly do this during initialization. I think together with this, it would be good to swap [these lines](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/b29869a64fbeb9eb9d09b14044ae85121e307feb/ur_robot_driver/resources/ros_control.urscript#L101-L102), as well.

ur_robot_driver/resources/ros_control.urscript
```diff
@@ -99,7 +99,7 @@ control_mode = MODE_UNINITIALIZED
 thread_move = 0
 global keepalive = -2
 textmsg("ExternalControl: External control active")
-params_mult = socket_read_binary_integer(1+6+1, "reverse_socket")
+params_mult = socket_read_binary_integer(1+6+1, "reverse_socket", 0)
```
fmauch, line 102: Those lines could be swapped so that the output only is printed if there was data received from the remote host.

giusebar: addressed in 21a4146ab8cb83dd8bd54e0abee670ee10e70428

fmauch:
Thank you @giusebar for your contribution.

Issue #306, opened 2021-01-13
Reading has_realtime fails sometimes


axelschroth:
I think this line should be changed from

https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/783b1d1f5774220058d0a2aa3dd93185641e8c7e/ur_robot_driver/src/hardware_interface_node.cpp#L60

to

`  bool has_realtime = false;`

I noticed that if I modify some unrelated code without initializing `has_realtime`, printing `has_realtime` sometimes outputs a number like e.g. 168 or 66. I suspect the ifstream to do nothing if `/sys/kernel/realtime` does not exist.
Any thoughts on that strange behavior?

fmauch:
Using the stream's `fail()` or `good()` functions might be reasonable. I've wanted to refactor this code for quite some time to also allow setting a higher thread priority if no realtime capability is found.

Initializing the variable and checking the ifstream might be a good first start to address the issue you mentioned.

Something like 
```c++
std::ifstream realtime_file("/sys/kernel/realtime", std::ios::in);
bool has_realtime = false;
if (realtime_file.good())
{
  realtime_file >> has_realtime;
}

@axelschroth could you maybe test whether this works for you?

**Comment by axelschroth** [comment]
Following works for me using `is_open()`:

```
std::ifstream realtime_file("/sys/kernel/realtime", std::ios::in);
bool has_realtime = false;
if (realtime_file.is_open())
{
  realtime_file >> has_realtime;
}
```

axelschroth:
Sorry for the confusion above. I created a PR.

fmauch:
This should have been fixed with #311 -> closing this issue.

PR #311, opened 2021-01-19
Fixes reading has_realtime property

axelschroth:
Reading this from system may end up in undefined behavior. (https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/issues/306)


fmauch (commented):
LGTM.

I'd like to test this with a machine WITH an RT-kernel and WITHOUT an RT kernel, but I don't expect anything to go wrong.

axelschroth:
Yes, tested WITHOUT RT-kernel only

fmauch:
@t-schnell Could you maybe arrange testing this on an RT kernel system?

fmauch:
I finally came across to test this on an RT system. Seems to work fine.

Thanks @axelschroth

PR #318, opened 2021-02-10
Use catkin_install_python macro for python files

mintar:
This macro works just like the normal `install` macro, but it also automatically changes the shebang line in the python file to `python2` or `python3`, depending which version is used.

See: http://wiki.ros.org/UsingPython3/SourceCodeChanges#Changing_shebangs

What this means is that this package can be used with Python3 without
any further changes, for example in ROS Noetic.


fmauch:
That sounds neat! As far as I read the source file this has actually been available since a long time so we shouldn't face any compatibility issues, right?

mintar:
Yes, I've tested it with kinetic, and it worked as expected.Am 10.02.2021 21:42 schrieb Felix Exner <notifications@github.com>:
That sounds neat! As far as I read the source file this has actually been available since a long time so we shouldn't face any compatibility issues, right?

fmauch:
Thank you for the contribution!

PR #330, opened 2021-03-08
Replace two logging macros with ROS logging macros.

fmauch:
Inside the driver we want to use plain ROS logging instead of the library's logging macros.


urrsk (approved):
Look good

urrsk:
Well found! @fmauch will you also cherry-pick or rebase thi change to the beta branch?

PR #331, opened 2021-03-09
Replace two logging macros with ROS logging macros. (#330)

fmauch:
Inside the driver we want to use plain ROS logging instead of the library's logging macros.


urrsk (approved):
Thanks

fmauch:
@stefanscherzinger I'll not merge this, as the beta-test is your area. However, please consider merging this soon, as this is holding back https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/40

PR #333, opened 2021-03-10
Added header and control loop definitions.

urmahp:
Defining header and control in the script enables the possibility to have multiple URCap nodes in one program. 
This relates to pr https://github.com/UniversalRobots/Universal_Robots_ExternalControl_URCap/pull/14

### DIFF
--- a/ur_robot_driver/resources/ros_control.urscript
+++ b/ur_robot_driver/resources/ros_control.urscript
@@ -1,5 +1,7 @@
 {{BEGIN_REPLACE}}
 
+# HEADER_BEGIN
+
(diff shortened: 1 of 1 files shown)


fmauch:
Let's move that over to the library as we want to move this script to the library, anyway.

urmahp:
> Let's move that over to the library as we want to move this script to the library, anyway.

Good idea I will open a PR on the client library with these changes instead, when the script has been moved.

fmauch:
Closing this in favor of #342 and https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/48

PR #342, opened 2021-03-22
Use external_control urcap version 1.0.5

fmauch:
As this version introduces the program separation into HEADER part and a control loop part, the script gets modified here, as well.


stefanscherzinger (approved):
Looks good.

Issue #325, opened 2021-02-28
Resetting infinite wrist_3_link [Question]


cambel:
# Summary

Hi, I am using the last link of the UR3e for a long rotation (several full rotations). For this, I move the robot using JointTrajectory messages which work fine. However, I am also using MoveIt for motion planning with obstacle avoidance, for which I set a finite joint limit to each joint. The problem is that if the joint position is outside the joint limits, MoveIt will complain about it so I have to manually reverse as many rotations, of the last link, to go within the valid range. Doing this reverse motion is way too inefficient.

Now, I notice that using the Polyscope GUI, I can set a target position for the **wrist_3**, for example, 0 degrees, then use the automatic option to move the robot. In that case, the robot will move less than a full revolution and reset the state of the robot to the target position.  

My question is: can this be done through this ROS driver? can I somehow request/reset the target position of the infinite rotation wrist_3_link via ROS?

# Versions
 - ROS version: Melodic / Ubuntu 18.04 (Docker)
 - ROS Driver version: 1.0.4

fmauch:
I think the short answer would be no.

But I'm not sure, whether there might be a workaround / solution for this.

The joint state is read by the hardware interface using the `actual_q` field from the [RTDE protocol](https://www.universal-robots.com/articles/ur/interface-communication/real-time-data-exchange-rtde-guide/). So, in order to "reset" the joint's value, we would have to either reset the joint state on the robot controller (which I am not sure whether it is possible), or substract / add a full revolution inside the driver and keep track of this, which I would not be such a huge fan of doing.

I guess, there are other robots out there facing this problem, so it would probably be a good idea to have a look how this situation is handled there.

fmauch:
Yes, that would be a manual workaround. If you found something, please post it here, so that 

 1. others can find this information, as well
 2. we can think whether and how we could integrate this into the driver.

cambel:
I found a workaround using URscript:

1. Since we have to use URScript, the connection to the robot e-series has to be initialized using `headless_mode`
2. Publish a custom script snippet to the topic `/ur_hardware_interface/script_command`
    the code is: `reset_revolution_counter()`

I understand the limitation of sending urscript code without being in headless_mode and this solution seems simple enough, but it would be nice to have a more direct option in the driver to do this operation.

fmauch:
Why would you require `headless_mode` for this to work?

cambel:
[...]

PR #343, opened 2021-03-24
Adding reset revolution counter service (#325)

cambel:
Hi,

I am trying to implement the service to reset the revolution counter of the robot's wrist_3_link. However, I am having a problem with the behavior of the script.

If I send the script as a secondary program:
```
sec resetRevolutionCounter():
  reset_revolution_counter()
end
```
It does not work as intended, the wrist_3_link gives a full rotation at full speed, which is not desirable at all, but on top of that, the counter is not reset. 

On the other hand, if I send the script just as:
```
reset_revolution_counter()
```

It does work as intended with the caveat that the driver's program is stopped and has to be reconnected in order to send any new command.

I think that the best solution would be to be able to reset the counter without disconnecting the main driver's program. Similarly to the behavior of the `zero_ftsensor` service. 

I would appreciate any suggestions on what I may be doing wrong.

### DIFF
--- a/ur_robot_driver/doc/ROS_INTERFACE.md
+++ b/ur_robot_driver/doc/ROS_INTERFACE.md
@@ -760,6 +760,10 @@ Setup the mounted payload through a ROS service
 
 Calling this service will zero the robot's ftsensor. Note: On e-Series robots this will only work when the robot is in remote-control mode.
 
+##### reset_revolution_counter ([std_srvs/Trigger](http://docs.ros.org/api/std_srvs/html/srv/Trigger.html))
+
+Calling this service will reset the revolution counter for the robot's wrist_3_link. Note: On e-Series robots this will only work when the robot is in remote-control mode.
+
 #### Parameters
 
 ##### dashboard/receive_timeout (Required)
--- a/ur_robot_driver/include/ur_robot_driver/hardware_interface.h
+++ b/ur_robot_driver/include/ur_robot_driver/hardware_interface.h
@@ -200,6 +200,7 @@ protected:
   bool setIO(ur_msgs::SetIORequest& req, ur_msgs::SetIOResponse& res);
   bool resendRobotProgram(std_srvs::TriggerRequest& req, std_srvs::TriggerResponse& res);
   bool zeroFTSensor(std_srvs::TriggerRequest& req, std_srvs::TriggerResponse& res);
+  bool resetRevolutionCounter(std_srvs::TriggerRequest& req, std_srvs::TriggerResponse& res);
   void commandCallback(const std_msgs::StringConstPtr& msg);
 
   std::unique_ptr<urcl::UrDriver> ur_driver_;
@@ -215,6 +216,7 @@ protected:
 
   ros::ServiceServer deactivate_srv_;
   ros::ServiceServer tare_sensor_srv_;
+  ros::ServiceServer reset_revolution_counter_srv_;
   ros::ServiceServer set_payload_srv_;
 
   hardware_interface::JointStateInterface js_interface_;
--- a/ur_robot_driver/src/hardware_interface.cpp
+++ b/ur_robot_driver/src/hardware_interface.cpp
@@ -384,6 +384,10 @@ bool HardwareInterface::init(ros::NodeHandle& root_nh, ros::NodeHandle& robot_hw
   // work when the robot is in remote-control mode.
   tare_sensor_srv_ = robot_hw_nh.advertiseService("zero_ftsensor", &HardwareInterface::zeroFTSensor, this);
 
+  // Calling this service will reset the revolution counter for the robot's wrist_3_link. Note: On e-Series robots this will only
+  // work when the robot is in remote-control mode.
+  reset_revolution_counter_srv_ = robot_hw_nh.advertiseService("reset_revolution_counter", &HardwareInterface::resetRevolutionCounter, this);
+
   ur_driver_->startRTDECommunication();
   ROS_INFO_STREAM_NAMED("hardware_interface", "Loaded ur_robot_driver hardware_interface");
 
@@ -909,6 +913,15 @@ bool HardwareInterface::zeroFTSensor(std_srvs::TriggerRequest& req, std_srvs::Tr
 end
 )");
   }
+  return true;
+}
+
+bool HardwareInterface::resetRevolutionCounter(std_srvs::TriggerRequest& req, std_srvs::TriggerResponse& res)
+{
+  res.success = this->ur_driver_->sendScript(R"(sec resetRevolutionCounter():
+  reset_revolution_counter()
+end
+)");
   return true;
 }
 


fmauch:
First of all, thanks for the PR!

Given your description the script function is indeed able to be run in a secondary function which would definitely be a good approach.

The problem is that the hardware interface [sends the current joint command](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/src/hardware_interface.cpp#L535) to the robot each control cycle. If that joint happens to be a full revolution off the current command, it will do a full rotation to get to the old configuration.

So, in order to prevent this, one could modify the `joint_position_command_` buffer to reflect the new configuration. This has multiple caveats:
 - As the control loop runs in a different thread than the service handler you'll have to deal with thread synchronization. And the control thread should not have to wait for anything...
 - Timing will be an issue here. When you send the `reset_revolution_counter()` to the robot you don't know exactly when this will actually happen. Any misalignment between setting the `joint_position_command_` buffer and the joint actually reporting being reset will result in a jerk in that joint.
 - What happens, if there is actually a command (trajectory goal) running when that happens. Modifying the output buffer will be overwritten by the trajectory almost instantaneously.

While the first two problems could probably be mitigated by doing numerous workarounds, the third one would actually remain a problem, but can also be seen as a hint to another possible solution.

If we shutdown all controllers actually moving the robot, then reset the counter and then restart the controller that has been running (maybe with a short delay to let the reset command do its job) the system will
 - not move due to the reset, as no joint commands will be sent, basically, we will end up in [this line](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/src/hardware_interface.cpp#L551) inside the `write()` function.
 - Any trajectories running at that point will be aborted and new trajectories will be rejected until the controller is running again.

The easiest way to achieve this would be to use the already running [controller stopper](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/master/controller_stopper/README.md). This could be done by simply publish a message to the `robot_program_running` topic as done [here](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/src/hardware_interface.cpp#L669-L671). Problem with that: This is meant to be only triggered from one source which is the driver's `ReverseInterface` that will check whether the corresponding URScript is active on the robot. As it will still remain active at that point, we are hijacking something not meant that way. We would have to find a way to clean that up properly.

However, as a proof of concept, this should work. While the reset is taking place, all "controlling" controllers (All except [those](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/launch/ur_control.launch#L76-L79)) should get stopped temporarily. The `control_mode` variable on the teach panel should become `0` during the operation and then go back to the value it was before.

cambel:
Thank you for the explanation!

I tried the approach that you mentioned, stopping the controllers > sending the script > restarting the controllers. I tested it and it worked. I had to add a delay after sending the script to allow the command to be processed before the controllers were back running.

fmauch:
OK, great! So now we need to find a way to make this less hacky and prune to errors.

My main concern is that this behavior is also controlled by [`HardwareInterface::handleRobotProgramState`](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/src/hardware_interface.cpp#L662).

One way around this might be to call this function directly instead of copying the contents and add a mutex to that function. This is necessary, as the service callback will be triggered by the [async spinner](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/src/hardware_interface_node.cpp#L50) while the changing program state will be triggered by the [Reverse interface watchdog](https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/master/src/ur/ur_driver.cpp#L204) (which runs in a separate thread).

Oh, and maybe it's about time to format your code according to the style guidelines :-) We use [clang-format](https://clang.llvm.org/docs/ClangFormat.html) in order to to that automatically. I usually use this script to format a complete repo:

```bash
#!/bin/bash
# This script uses clang-format to format all *.cpp, *.hpp and *.h files recursively

clang_format_exec="clang-format"

find . -name "*.cpp" -exec sh -c "echo 'Formatting {}' && $clang_format_exec -i {}" \;
find . -name "*.hpp" -exec sh -c "echo 'Formatting {}' && $clang_format_exec -i {}" \;
find . -name "*.h" -exec sh -c "echo 'Formatting {}' && $clang_format_exec -i {}" \;
```

cambel:
Thank you for the formatting script!

I made the suggested changes, I am not sure if I added the `mutex` as intended though. I tested it and it works. 

This is the behavior of the function when tested:
- If the angle of the wrist3 is bigger than 2*PI or less than -2*PI: The wrist3 moves slightly to some position and the revolution counter is reset. Though I am not sure why it moves.
- If the angle of the wrist3 is within [-2*PI, 2*PI]: The wrist3 does not move and the revolution counter does not change.

ur_robot_driver/doc/ROS_INTERFACE.md
```diff
@@ -760,6 +760,10 @@ Setup the mounted payload through a ROS service
 
 Calling this service will zero the robot's ftsensor. Note: On e-Series robots this will only work when the robot is in remote-control mode.
 
+##### reset_revolution_counter ([std_srvs/Trigger](http://docs.ros.org/api/std_srvs/html/srv/Trigger.html))
+
+Calling this service will reset the revolution counter for the robot's wrist_3_link. Note: On e-Series robots this will only work when the robot is in remote-control mode.
```
urrsk, line 765: Just want to clarify for the future reader that it is only the UR3 that will benefits from this.
```suggestion
The UR3 and UR3e's wrist 3 can do infinite rotations and are therefore not limited to ±2Pi.
Calling this service will reset the revolution counter for the robot's wrist_3_link. Note: On e-Series robots this will only work when the robot is in remote-control mode.
This is handy to avoid rotating back after e.g. a screwing task, and thereby avoid compensating for this in the program.
```

urrsk (commented):
>     * If the angle of the wrist3 is bigger than 2_PI or less than -2_PI: The wrist3 moves slightly to some position and the revolution counter is reset. Though I am not sure why it moves.

Thanks for that insight, I will try to get it verified

ur_robot_driver/doc/ROS_INTERFACE.md
gavanderhoorn: Might be nice to make this even stronger and have the service fail if the robot is not a UR3.

cambel: I suppose something like this:
```c++
bool HardwareInterface::resetRevolutionCounter(std_srvs::TriggerRequest& req, std_srvs::TriggerResponse& res)
{
  if (ur_driver_->getName() == 'UR3e' || ur_driver_->getName() == 'UR3')
  {
    std::mutex mtx;
    handleRobotProgramState(false);
    mtx.lock();
    res.success = this->ur_driver_->sendScript(R"(sec resetRevolutionCounter():
    reset_revolution_counter()
  end
  )");
    mtx.unlock();
    handleRobotProgramState(true);
  }
  else
  {
    std::stringstream ss;
    ss << "Resetting the revolution counter is only available for UR3/e robots. This robot is "
       << ur_driver_->getName();
    ROS_ERROR_STREAM(ss.str());
    res.message = ss.str();
    res.success = false;
  }
  return true;
}
```

Though, I don't know where to get the robot's name...

fmauch: Out of my head I am not sure whether there's something predefined to get the robot model through the API. But that would have been my next point on the list.

@gavanderhoorn wrote:
> Might be nice to make this even stronger and have the service fail if the robot is not a UR3.

I would even have supposed to advertise it only for UR3(e) robots. But that might be more confusing than simply letting the service fail...

fmauch:
> > * If the angle of the wrist3 is bigger than 2_PI or less than -2_PI: The wrist3 moves slightly to some position and the revolution counter is reset. Though I am not sure why it moves.
> 
> Thanks for that insight, I will try to get it verified

@cambel Just to make sure: Is this independent of using with the ROS driver?

cambel:
I am not sure. I've only tested it using the ROS driver.
I will do some tests next week using just plain sockets connection. 
What I know is that the slight motion of the wrist3 is not always happening, though today it happened all the time.

cambel:
Okay, I did some more tests.

Using the **URScript** function `resetRevolutionCounter()` do not move the robot, it only resets the revolution counter when the angle of the wrist3 is bigger than 2_PI or less than -2_PI.

The slight motion that I have been observing may be due to the slight delay added to the function `ros::Duration(0.1).sleep();`. I supposed that the controllers are not stopped before the URscript is called so the revolution counter is reset, then the controller tries to go to its latest know configuration for a slipt second before the controllers are stopped and resumed.

The problem is as @fmauch explained above, about the timing of the threads. I tried to implement the `mutex` but I've been unsuccessful to make it work properly (I have no experience doing that). No matter what I do, it does not seem to lock the resources so that the URscript function is executed only when the controllers are stopped and before they are resumed.

I'd appreciate some guidance on this part.

fmauch:
I'd like to nail things down to the actual problem:

Could you add sleeps (as in 1 sec) before [exclusive] or after calling the URScript function? Is the problem that the controllers aren't going down fast enough or coming up too early?

The problem with our current setup is that we have a (kind of hacky) fire-and-forget interface to stopping & starting the controllers. If we can nail down the problem, we can think of a more proper solution.

cambel:
@fmauch I will do more tests to try to understand how the driver is trying to execute the stopping & starting the controllers.

However, last time I did already try adding sleep of 1 second before/after calling the URScript, the result was the spin behavior I mentioned in the first commend of this thread. I even try only stopping the controller and then executing the URScript, but it goes back to working sometimes only when the functions sync properly I guess. After that, the controllers were running again without explicitly starting them again.

urrsk:
Maybe the best solution to do this is to subtract an angle offset in the driver, instead of trying to make a async. call sync. Or add it to the sync protocol.

PR #345, opened 2021-03-25
Take joint order into consideration for trajectory passthrough

stefanscherzinger:
This allows an arbitrary joint order in the trajectory goals and covers MoveIt's alphabetical ordering.

I have the following reasons why this should go into the driver:
- The rationale of the PassThroughControllers is to forward a trajectory representation unchanged to the RobotHW. Each driver should then try to implement with a _best effort_ policy whatever is most suitable for each robot.
- The driver has a knowledge about supported joints and their natural order in the robot's kinematics

### DIFF
--- a/ur_robot_driver/src/hardware_interface.cpp
+++ b/ur_robot_driver/src/hardware_interface.cpp
@@ -1186,20 +1186,50 @@ bool HardwareInterface::checkControllerClaims(const std::set<std::string>& claim
 
 void HardwareInterface::startJointInterpolation(const hardware_interface::JointTrajectory& trajectory)
 {
+  // Take joint order into consideration by
+  // computing the map between msg indices to expected indices.
+  // If msg is {C, B} and expected is {A, B, C, D}, the associated mapping vector is {2, 1}
+  //
+  auto msg = trajectory.trajectory.joint_names;
+  auto expected = joint_names_;
+  std::vector<size_t> msg_joints(msg.size());
+  if (msg.size() > expected.size())
+  {
+    // msg must be a subset of expected
+    LOG_WARN("Not forwarding trajectory. It contains too many joints");
+    return;
+  }
+  for (auto msg_it = msg.begin(); msg_it != msg.end(); ++msg_it)
+  {
+    auto expected_it = std::find(expected.begin(), expected.end(), *msg_it);
+    if (expected.end() == expected_it)
+    {
+      LOG_WARN("Not forwarding trajectory. It does not contain the expected joints");
+      return;
+    }
+    else
+    {
+      const size_t msg_dist = std::distance(msg.begin(), msg_it);
+      const size_t expected_dist = std::distance(expected.begin(), expected_it);
+      msg_joints[msg_dist] = expected_dist;
+    }
+  }
+
   size_t point_number = trajectory.trajectory.points.size();
   LOG_DEBUG("Starting joint-based trajectory forward");
   ur_driver_->writeTrajectoryControlMessage(urcl::comm::TrajectoryControlMessage::TRAJECTORY_START, point_number);
   double last_time = 0.0;
-  for (size_t i = 0; i < point_number; i++)
+  for (auto point : trajectory.trajectory.points)
   {
-    trajectory_msgs::JointTrajectoryPoint point = trajectory.trajectory.points[i];
-    urcl::vector6d_t p;
-    p[0] = point.positions[0];
-    p[1] = point.positions[1];
-    p[2] = point.positions[2];
-    p[3] = point.positions[3];
-    p[4] = point.positions[4];
-    p[5] = point.positions[5];
+    // Only write to joints that are contained in the new point.
+    // Keep the other joints unchanged.
+    urcl::vector6d_t p = joint_positions_;
+    for (size_t i = 0; i < msg_joints.size(); ++i)
+    {
+      auto jnt_id = msg_joints[i];
+      p[jnt_id] = point.positions[i];
+    }
+
     double next_time = point.time_from_start.toSec();
     ur_driver_->writeTrajectoryPoint(p, false, next_time - last_time);
     last_time = next_time;


gavanderhoorn:
I was always under the impression the order of joints as specified in the `joints` (or `joint_names`) parameter would be what maps indices to names.

Controllers then access `JointHandle`s using those names, thus never having to deal with indices.

In this PR, it looks like the hw interface has to shuffle things around again, but I'm not sure why the hw iface would ever see anything other than writes to handles?

stefanscherzinger:
> Edit: o wait, this is caused by the fact that the new implementation of the hw iface (for the delegate-interpolation-to-real-hw feature) implements a bit of a trick, correct, and thereby breaks the normal flow of information (and associated assumptions)?

Yes, you are right. Normal ROS controllers should use the handles as unique identifiers. The new hardware interface for trajectory forwarding is however different to this approach. We needed some trajectory buffer for this.

I'm thinking now if it makes sense to re-order the trajectory representation in the PassThroughController. We need the joint names there anyway for resource claiming. The caveat that's left would be that users could specify a different joint order for their PassThroughController that would then clash with the driver's kinematics.

Am I right that users should decide on the resources they wish to use by specifying them for each controller? Or is it _ok_ to retrieve this information somehow from the RobotHW?

gavanderhoorn:
> I'm thinking now if it makes sense to re-order the trajectory representation in the PassThroughController. 

that would be a nicer way of solving this I believe, yes.

> The caveat that's left would be that users could specify a different joint order for their PassThroughController that would then clash with the driver's kinematics.

That's true. I wouldn't call it *kinematics* though. You mean "the driver's joint order", correct?

> Am I right that users should decide on the resources they wish to use by specifying them for each controller? 

yes.

> Or is it ok to retrieve this information somehow from the RobotHW?

no. Not normally.

It's possible though: the `joint_state_controller` does this (from [here](https://github.com/ros-controls/ros_controllers/blob/a075bc36ebf5b81a7dafe911907cc1b781dd0185/joint_state_controller/src/joint_state_controller.cpp#L45-L49)):

```c++
  bool JointStateController::init(hardware_interface::JointStateInterface* hw,
                                  ros::NodeHandle&                         root_nh,
                                  ros::NodeHandle&                         controller_nh)
  {
    // get all joint names from the hardware interface
    const std::vector<std::string>& joint_names = hw->getNames();
    num_hw_joints_ = joint_names.size();
    for (unsigned i=0; i<num_hw_joints_; i++)
      ROS_DEBUG("Got joint %s", joint_names[i].c_str());

    [...]
```

it works for that controller as it doesn't claim the resources exclusively.

I'm not sure how that would work for actual *controlling* controllers, or with `CombinedRobotHW`.

gavanderhoorn:
> If we got the _driver's joint order_ somehow from the RobotHW, then we could take the subset as specified for the PassThroughController and do the re-ordering there. If possible, that would be the preferred way so that other drivers need not re-implement this themselves.

yes, this would be the best possible way to implement this.

Anything is would be *very* brittle.

> Any chance that
> 
> ```c++
> const std::vector<std::string>& joint_names = hw->getNames();
> ```
> 
> already provides the driver's _natural_ joint order (in subsets)?

No, I would not expect that to be the case.

Most drivers / `RobotHW` implementations do either of these two things:

 1. create joint handles in whatever order the user has configured as part of a `joints` or `joint_names` parameter
 2. create joint handles in *some* order regardless of what a user may have configured. That order does not necessarily *have* to be any sort of "natural order". It could be. But it doesn't have to be

Personally I use 2 for many of the `RobotHW` implementations I've created. But that's primarily because I prefer to write drivers which use the external hw as much as possible for information on existence of resources (and their characteristics). If the robot controller already has the absolute truth (ie: knows which resources it exposes, and in which order (for joints fi)), there's nothing I can do about it anyway, so just use whatever the controller tells me.

The problem with both options is that you cannot really *assume* anything in controllers (and normally you would never, this passthrough controller is a special case).

I'm not sure I can think of an elegant way to solve this generically (ie: without having to resort to methods outside of regular `ros_control`).

stefanscherzinger:
> The problem with both options is that you cannot really assume anything in controllers (and normally you would never, this passthrough controller is a special case).
I'm not sure I can think of an elegant way to solve this generically (ie: without having to resort to methods outside of regular ros_control).

That's true. Some robot-driver combinations could provide this, but it still needs to be generic for ROS control.

That's why this driver currently _trusts_ its according [configuration on the parameter server](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/src/hardware_interface.cpp#L289). I don't see any sanity check with the UR robot's _natural order_.

The _joint_trajectory_controller_ also goes this way: It uses the [user-specified order](https://github.com/ros-controls/ros_controllers/blob/noetic-devel/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L242) and takes that to overrule the message order ([here](https://github.com/ros-controls/ros_controllers/blob/noetic-devel/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L522) and [here](https://github.com/ros-controls/ros_controllers/blob/noetic-devel/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L532)).

So, users of MoveIt and the joint_trajectory_controller (or the scaled_joint_trajectory_controller) need to make estimated guesses on the _natural order of joints_ for their robot. After all, users specify controllers for _their robot_, so I guess it makes sense to tolerate this.

Following these examples, the PassThroughControllers would use [their specified joints](https://github.com/fzi-forschungszentrum-informatik/cartesian_ros_control/blob/beta-testing/pass_through_controllers/examples/hw_interface/controllers.yaml#L22) and perform the reordering according to that.

stefanscherzinger:
Hm, two things just came to my mind:

> Anything else would be very brittle (as it delegates this responsibility to RobotHW implementations which most likely have not done this, as this controller didn't exist when it was written, which reduces reusability of the passthrough controller significantly).

Unfortunately, implementers of RobotHW need to change their implementation anyway, because they need the new trajectory_interface mechanism.

Currently, this PR allows users to specify subsets of joints for their trajectory while forwarding. This is possible, because the driver can read joint positions and keep the unspecified ones where they are. This would get lost when the PassThroughController implement joint reordering, because partially setting joint commands does not make sense on their abstraction level.

To facilitate things for RobotHW implementers, we could provide the joint re-ordering functionality as part of the pass_through_controller library (along with the new  trajectory_interface).

fmauch (commented):
I would vote for moving the reordering to the controller, as it seems tedious to implement this in every hardware interface.

With the implementation given here, it is implicitly expected, that the `joint_names` parameter keeps the joints in the correct order, which is also the assumption for the joint position and joint velocity controllers.

The least we could do is to factor this out into a method offered by the passthrough controller that takes the expected joint order as a parameter. However, as this is not changing during runtime, we could add this to the initialization process, e.g. when creating the handle.
One benefit of that would be that we can verify the controller configuration with regards that the relevant joints are in fact a subset of the expected joints.

stefanscherzinger:
@fmauch @gavanderhoorn Here's [an alternative suggestion](https://github.com/fzi-forschungszentrum-informatik/cartesian_ros_control/pull/24) that does the reordering on the PassThroughController side.

We will lose the feature to send joint trajectories, though, where only one joint is specified (e.g. the last joint for rotation) and the other joints are intended to stay where they are.

PR #88, opened 2020-01-24
Parameterising gains

carebare47:
Added two parameters for servoj_gain and servoj_lookahead_time

PR #354, opened 2021-04-22
add arg for servoj_gain servoj_lookahead_time in ur_control.launch

lianghongzhuo:
as the pull request https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/88/files
added the parameter to set servoj_gain and servoj_lookahead_time, add these parameters also in the launch file.


fmauch (approved):
I would not like to add it to the toplevel launchfiles to not blow up the parameter interface for new users. However, only adding this to the `ur_control.launch` file seems fine to me.

lianghongzhuo:
Thanks, we need it as our ur5 setup wants a smoother motion, and we include `ur_control.launch` directly. So it would be easier if this file contains such parameters.  [link to our launch file](https://github.com/TAMS-Group/tams_ur5_setup/blob/master/tams_ur5_setup_bringup/launch/tams_ur5_setup_ur_robot_driver.launch#L10-L31)

fmauch:
Thanks @lianghongzhuo !

PR #355, opened 2021-04-23
Optionally load robot description

christianlandgraf:
Hello,

I would like to propose a parameter to optionally load the robot_description  in ```ur_common.launch```.
I'm not completely sure if it is necessary, but it might help to prevent confusion if ```robot_description``` has been loaded previously or somewhere else with custom dh configuration. Other robot drivers have this option as well (e.g. [Pilz](https://github.com/PilzDE/pilz_robots/blob/noetic-devel/prbt_support/launch/robot.launch) )


fmauch:
Would it make sense to not load the description, but start the `robot_state_publisher`? Usually (in my experience), that's done together. If you don't want to load the description and start the `robot_state_publisher`, users can use the `ur_control.launch` file directly, at least that was our intention.

christianlandgraf:
I agree, that's a better idea :+1:

PR #366, opened 2021-05-10
Call calibration check in ROS driver

fmauch:
Mandatory check in the client library has been deprecated.

This uses the changes introduced in https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/65

This should be merged after ur_client_library has been binary-released with this change.


fmauch:
Merging to staging as pipeline is successful for the testing repo

PR #381, opened 2021-05-20
Robot mode in set mode

fmauch:
This should make the message `SetMode` action easier to interpret. My initial motivation was that the mode itself is documented inside the `RobotMode.msg` file.


RobertWilbrandt (approved):
Seems reasonable to do this before the release. Thanks!

PR #350, opened 2021-04-09
add reserve_ip

JS00000:
This PR is copied from [here](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/268), which can help the socket in the UR's Teach Pendant know the host IP address when this library is running in the Docker container (especially in Docker for Windows).

PR #412, opened 2021-06-22
Reverse ip

fmauch:
Rebased version of #350


fmauch:
Merging, as all jobs with the testing repo succeed.

Thanks @JS00000 for sticking with this!

PR #413, opened 2021-06-23
Cartesian control

fmauch:
This PR adds Cartesian streaming control and controllers to the driver.

In analogy to the JointPosition and JointVelocity interface, a Pose and Twist interface is added. We have three controllers added as  examples:
 - `twist_controller`: publish a cartesian twist to the robot's tcp
 - `pose_based_cartesian_traj_controller`: Cartesian trajectory execution using a pose interface
 - `joint_based_cartesian_traj_controller`: Cartesian trajectory execution using an IK and a joint position interface

### DIFF
--- a/ur_robot_driver/package.xml
+++ b/ur_robot_driver/package.xml
@@ -41,10 +41,12 @@
   <depend>tf2_msgs</depend>
   <depend>tf</depend>
   <depend>trajectory_msgs</depend>
+  <depend>twist_controller</depend>
   <depend>ur_client_library</depend>
   <depend>ur_dashboard_msgs</depend>
   <depend>ur_msgs</depend>
 
+  <exec_depend>cartesian_trajectory_controller</exec_depend>
   <exec_depend>controller_stopper</exec_depend>
   <exec_depend>force_torque_sensor_controller</exec_depend>
   <exec_depend>industrial_robot_status_controller</exec_depend>
--- a/ur_robot_driver/package.xml
+++ b/ur_robot_driver/package.xml
@@ -41,7 +41,6 @@
   <depend>tf2_msgs</depend>
   <depend>tf</depend>
   <depend>trajectory_msgs</depend>
-  <depend>twist_controller</depend>
   <depend>ur_client_library</depend>
   <depend>ur_dashboard_msgs</depend>
   <depend>ur_msgs</depend>
@@ -54,6 +53,7 @@
   <exec_depend>joint_trajectory_controller</exec_depend>
   <exec_depend>robot_state_publisher</exec_depend>
   <exec_depend>socat</exec_depend>
+  <exec_depend>twist_controller</exec_depend>
   <exec_depend>ur_description</exec_depend>
   <exec_depend>velocity_controllers</exec_depend>
 
--- a/ur_robot_driver/src/hardware_interface.cpp
+++ b/ur_robot_driver/src/hardware_interface.cpp
@@ -799,7 +799,6 @@ void HardwareInterface::doSwitch(const std::list<hardware_interface::ControllerI
         }
         if (resource_it.hardware_interface == "ros_controllers_cartesian::PoseCommandInterface")
         {
-          ROS_INFO_STREAM("Stopping pose controller");
           pose_controller_running_ = false;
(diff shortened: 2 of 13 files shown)


ur_robot_driver/package.xml
```diff
@@ -41,10 +41,12 @@
   <depend>tf2_msgs</depend>
   <depend>tf</depend>
   <depend>trajectory_msgs</depend>
+  <depend>twist_controller</depend>
```
stefanscherzinger, line 44: Would `exec_depend` be sufficient here?

ur_robot_driver/src/hardware_interface.cpp
```diff
@@ -746,6 +793,15 @@ void HardwareInterface::doSwitch(const std::list<hardware_interface::ControllerI
         {
           cartesian_forward_controller_running_ = false;
         }
+        if (resource_it.hardware_interface == "ros_controllers_cartesian::TwistCommandInterface")
+        {
+          twist_controller_running_ = false;
+        }
+        if (resource_it.hardware_interface == "ros_controllers_cartesian::PoseCommandInterface")
+        {
+          ROS_INFO_STREAM("Stopping pose controller");
```
stefanscherzinger, line 802: This seems to be the only controller type that gets an info output. Is this intention?

ur_robot_driver/src/hardware_interface.cpp
```diff
@@ -784,6 +840,15 @@ void HardwareInterface::doSwitch(const std::list<hardware_interface::ControllerI
         {
           cartesian_forward_controller_running_ = true;
         }
+        if (resource_it.hardware_interface == "ros_controllers_cartesian::TwistCommandInterface")
+        {
+          twist_controller_running_ = true;
+        }
+        if (resource_it.hardware_interface == "ros_controllers_cartesian::PoseCommandInterface")
+        {
+          ROS_INFO_STREAM("Starting pose controller");
```
stefanscherzinger, line 849: Same here... Output wanted?

ur_robot_driver/test/integration_test.py
```diff
@@ -87,6 +91,15 @@ def init_robot(self):
                 "Could not reach joint passthrough controller action. Make sure that the driver is actually running."
                 " Msg: {}".format(err))
 
+        self.cartesian_trajectory_client = actionlib.SimpleActionClient(
+            'follow_cartesian_trajectory', FollowCartesianTrajectoryAction)
+        try:
```
stefanscherzinger, line 96: I guess `wait_for_server` does not throw but [returns false on timeout](https://docs.ros.org/en/noetic/api/actionlib/html/simple__action__client_8py_source.html#l00067).
Maybe simply use:

```python
        if not self.cartesian_trajectory_client.wait_for_server(timeout):
            self.fail("Could not reach cartesian controller action. Make sure that the driver is actually running.")

```

stefanscherzinger (approved):
Looks correct to me. Left some minor suggestions.

ur_robot_driver/package.xml
fmauch: yes, that should be sufficient

ur_robot_driver/src/hardware_interface.cpp
fmauch: That seems to be a leftover from development. I will remove that.

ur_robot_driver/src/hardware_interface.cpp
fmauch: Same as above

ur_robot_driver/test/integration_test.py
fmauch: Thanks for spotting this. `ServiceProxy.wait_for_service` [throws an exception](https://github.com/ros/ros_comm/blob/noetic-devel/clients/rospy/src/rospy/impl/tcpros_service.py#L93), while  [`ActionClient.wait_for_server`](https://docs.ros.org/en/noetic/api/actionlib/html/action__client_8py_source.html#l00590) does not.

fmauch: Resolved in 65e55ac

PR #438, opened 2021-07-21
Update README.md

pheobeyeung:
update download for rqt controller


fmauch (approved):
Thanks for the input @pheobeyeung

PR #446, opened 2021-08-03
Publish joint temperature

tassos:
Since we need temperature information on the motor level for parameters identification, I implemented publishers for publishing the joint temperature as retrieved by the driver. I am making a pull request in case this would be useful for others too, and to get some feedback on my implementation.

Based on previous discussions at #35 and ros-industrial/ur_modern_driver#81, I used the `sensor_msgs::Temperature` message type and constructed a publisher for each joint. I have two questions:

* Would it make sense to create a custom message for publishing an array of temperatures, so that only one topic would be enough?
* I read at ros-industrial/ur_modern_driver#315 and ros-industrial/ur_modern_driver#320 that link names might be preferred over joint names for the `frame_id`. Is this indeed the case? What would be the most convenient way to retrieve link names?

### DIFF
--- a/ur_robot_driver/include/ur_robot_driver/hardware_interface.h
+++ b/ur_robot_driver/include/ur_robot_driver/hardware_interface.h
@@ -259,7 +259,8 @@ protected:
   std::bitset<11> safety_status_bits_;
 
   std::unique_ptr<realtime_tools::RealtimePublisher<tf2_msgs::TFMessage>> tcp_pose_pub_;
-  std::unique_ptr<realtime_tools::RealtimePublisher<sensor_msgs::Temperature>> joint_temperatures_pub_;
+  typedef std::shared_ptr<realtime_tools::RealtimePublisher<sensor_msgs::Temperature>> JTPublisherPtr;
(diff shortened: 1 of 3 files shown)


fmauch:
Please see #35

I'd like to get this fixed all together.

tassos:
I can understand that this PR is way more general, but is it actually still work in progress, cause it seems a bit stalled.

fmauch:
That's right. That also means, that any help in this would be very appreciated :-)

tassos:
Ok got it. I will see if I can somehow contribute there then. Feel free to close this pull request then, and in case someone needs urgently temperature measurements, they can check out the corresponding branch.

fmauch:
> Ok got it. I will see if I can somehow contribute there then. Feel free to close this pull request then, and in case someone needs urgently temperature measurements, they can check out the corresponding branch.

That's why I would leave it open. It would be awesome if you could contribute to #35 I would very much like to get this done, but couldn't find the time to tackle it, yet.

fmauch:
please keep.

Issue #447, opened 2021-08-06
In the hardware_interface, some members should be atomic


AndyZe:
It seems like `robot_program_running_` and `controller_reset_necessary_` should be std::atomic<bool> type for thread safety. I say this because it looks like there is potential to read the variables at the same time as writing them from a callback.

Maybe others as well, like `position_controller_running_` and `velocity_controller_running_`

fmauch:
Thanks for bringing this up. We will definitely look into this!

fmauch:
This  has been fixed in #448 -> closing

PR #448, opened 2021-08-06
Make several members of hw_interface atomic, for thread safety

AndyZe:
Addresses #447.

I look at this as cheap insurance because there's no downside to making these variables atomic AFAIK.


stefanscherzinger:
Thank you @AndyZe for this contribution! I'll have a look.

fmauch (approved):
As I like to use PRs to learn something, please let me ask to elaborate this a bit.

I get that using atomics for `robot_program_running_` and `controller_reset_necessary_` is clean (although from my understanding on Linux the set and get operations should be atomic anyway). They are used inside a callback function which is triggered from the `ReverseInterface` thread.

`position_controller_running_` and `velocity_controller_running_` is relevant, as `doSwitch` is triggered outside of the usual ROS callback mechanisms, but in the real-time ROS-Control thread, right?

Skimming over the functions of our HardwareInterface this seems to cover all relevant members from my point of view. Thanks @AndyZe for addressing this and making a PR.

AndyZe:
@fmauch  you probably know more about it than I do. The way I look at it is:

`position_controller_running_` and `velocity_controller_running_` are used in write() and doSwitch(). They are probably threadsafe if ros_control does its job of stopping the read/write loop before switching. So I could probably revert those if you want me to.

`robot_program_running_` is used in write(), isRobotProgramRunning(), handleRobotProgramState(), and stopControl(). That's quite a few places that I don't all trust to be threadsafe (but maybe they are...)

Same thing with `controller_reset_necessary_`. It's used in read(), handleRobotProgramState(), and shouldResetControllers(). Several places that might not all be threadsafe.

fmauch:
> `position_controller_running_` and `velocity_controller_running_` are used in write() and doSwitch(). They are probably safe if ros_control does its job of stopping the read/write loop before switching. So I could probably revert those if you want me to.

No, I'm absolutely fine with keeping them. @stefanscherzinger do you think this would make sense to merge into master or staging? Intuitively I would merge it into master, but it will probably create conflicts with staging. Additionally, we introduced other members in the staging branch that should get the same treatment. I would go for merging this into master and adapting staging accordingly.

fmauch:
As we plan to merge `staging` very soon, I'd postpone this after the merge.

fmauch:
Looking at the conflicts it looks as if we have to extend this PR with the new interfaces. @AndyZe would you like to do it or should I update this PR?

Issue #272, opened 2020-09-29
Support for booting UR-computer after ROS-driver


jorgenfb:
I would like the driver to support waiting for the UR computer to boot up.

The driver will try to connect, but exit if unable to do so within a reasonable time. Since the ur_robot_driver is marked as required in the launch file the entire ROS system will be brought down with it. 

I would like the driver to support waiting indefinitely for the robot to become available. This would allow us to instruct our users to turn on the robot if its not available. Todays solution just kills everything, including our user interface support systems.

I would like your opinion on this. Is this something you find useful and will consider supporting?

fmauch:
The problem actually goes a bit deeper than that... If the driver gets started before the UR-controller is actually booted, things get bad as some subsystems are restarted again. As far as I remember, the RTDE handshake produced an exception, as it was restarted at some point during initialization. I don't remember it in detail, though.

And yes, that would definitely be nice to have, it was just never considered important enough to get ontop of the priority list. If you'd like to work on this, please let us know.

jorgenfb:
I will spend some time working on this. For my use it is crucial that the whole ROS system is not brought down if the communication with the UR bot fails on boot.

I have tried digging into the details in the source. And the chain of initialization for the ur_hardware_interface is something along:

```
ur_hardware_interface -> ur_driver (with its socket connections) -> rtde_client
```

There are multiple potensial failures in that chain. Especially if trying to connect while the UR-machine is booting. I am hoping to not touch all of those areas and instead opt for just allowing the ur_hardware_interface node to die and restart to try again.

Allowing the ur_hardware_interface to die leads to issues with controllers starting to early. This seems to be solvable by having the control_manager wait until connection is established. The control_manager already has support for waiting until a std_msgs/Bool topic emits true. We could create this topic in the ur_hardware_interface to signal that the driver is ready. Adding this ready signal to the driver would also be useful in userland applications (like ours) to determine when the robot driver is good to go.

Proposal for the ready topic:
Topic name: `/ur_hardware_interface/ready`
Topic type: `std_msgs/Book`
Topic behavior: Initially emit false until a message on both `/ur_hardware_interface/robot_mode` and `/ur_hardware_interface/safety_mode` has been emitted.

Does this sound like a feasible approach?

fmauch:
In my opinion this seems like a workaround for some use cases, but not a proper solution.

You are right, that the problem is twofold:
[...]

PR #450, opened 2021-08-12
Make hw_interface-node required-argument optional

steinmn:
The UR-robot is only one part of our roslaunch-setup, so I would like to be able to have the rest of the system (non-UR) continue to run even if the ur_hardware_interface-node dies. This is primarily to be able to inspect and collect diagnostics before displaying the appropriate information and available options (most likely restarting ros) to the user.

### DIFF
--- a/ur_robot_driver/launch/ur_control.launch
+++ b/ur_robot_driver/launch/ur_control.launch
@@ -31,9 +31,10 @@
   <arg name="headless_mode" default="false" doc="Automatically send URScript to robot to execute. On e-Series this does require the robot to be in 'remote-control' mode. With this, the URCap is not needed on the robot."/>
   <arg name="servoj_gain" default="2000" doc="Specify gain for servoing to position in joint space. A higher gain can sharpen the trajectory."/>
   <arg name="servoj_lookahead_time" default="0.03" doc="Specify lookahead time for servoing to position in joint space. A longer lookahead time can smooth the trajectory."/>
+  <arg name="ur_hardware_interface_node_required" default="true" doc="Shut down ros environment if ur_hardware_interface-node dies."/>
 
   <!-- Load hardware interface -->
-  <node name="ur_hardware_interface" pkg="ur_robot_driver" type="ur_robot_driver_node" output="screen" launch-prefix="$(arg launch_prefix)" required="true">
+  <node name="ur_hardware_interface" pkg="ur_robot_driver" type="ur_robot_driver_node" output="screen" launch-prefix="$(arg launch_prefix)" required="$(arg ur_hardware_interface_node_required)">
     <param name="robot_ip" type="str" value="$(arg robot_ip)"/>
     <param name="reverse_ip" value="$(arg reverse_ip)"/>
     <param name="reverse_port" type="int" value="$(arg reverse_port)"/>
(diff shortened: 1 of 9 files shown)


stefanscherzinger:
Might be related to #272

ur_robot_driver/launch/ur_control.launch
```diff
@@ -29,9 +29,11 @@
...
   <arg name="servoj_gain" default="2000" doc="Specify gain for servoing to position in joint space. A higher gain can sharpen the trajectory."/>
   <arg name="servoj_lookahead_time" default="0.03" doc="Specify lookahead time for servoing to position in joint space. A longer lookahead time can smooth the trajectory."/>
+  <arg name="ur_hardware_interface_node_required" default="true" doc="Shut down ros environment if ur_hardware_interface-node dies."/>
+  <arg name="ur_hardware_interface_node_respawn" default="false" doc="Restart (after 30 second delay) ur_hardware_interface-node if it dies."/>
 
   <!-- Load hardware interface -->
-  <node name="ur_hardware_interface" pkg="ur_robot_driver" type="ur_robot_driver_node" output="screen" launch-prefix="$(arg launch_prefix)" required="true">
+  <node name="ur_hardware_interface" pkg="ur_robot_driver" type="ur_robot_driver_node" output="screen" launch-prefix="$(arg launch_prefix)" required="$(arg ur_hardware_interface_node_required)" respawn="$(arg ur_hardware_interface_node_respawn)" respawn_delay="30">
```
fmauch, line 36: Instead of hard-coding a delay and providing a boolean flag, we could provide the respawn delay directly.

fmauch (commented):
In principle I fully agree to this PR (Though I think providing the respawn delay directly would be better, see my other comment).

However, adding a respawn delay feels not right to me. If we expect something to crash so we'll have to respawn it, we should put our energy into fixing the crashes instead. @steinmn why would you see the need for a respawn flag?

Concerning the required flag I absolutely agree it makes sense to move this to the top-level launchfiles.

steinmn:
I fully agree that the ideal solution is to fix all potential crashes, but the potential for some new and unexpected way of failing is in my experience always present in a sufficiently complex system. While respawning might not recover the full functionality, it could potentially recover functionality useful for debugging and/or softer shutdown.

Using the situation that triggered me making this PR as an example:
The node dying seems to have been triggered by a C193A7-error in the UR-system (error in the safety processor of the teach pendant), leading to the following:
`Jun 21 07:45:30 [753]: terminate called after throwing an instance of 'ur_driver::UrException'`
`Jun 21 07:45:30 [753]:   what():  No answer to RTDE protocol version negotiation request was received from robot. This should not happen!`
This error is pretty much catastrophic, so the only way to fully fix it is to reboot the ur-pc anyway. However, it would be very useful to extract the service-file/flight-reports for debugging later, which I use the `/ur_hardware_interface/dashboard/raw_request`-service for.

Now I have no idea if this service would have been functional if we had respawned the node in the example case, since I don't know of any way to reproduce the issue. I have tested just killing and respawning the node on an otherwise functional system, and both dashboard-services and io-services seemed to work, while no joint_state-messages were being sent. To me at least, this partial recovery still represents some value in terms of debugging.

fmauch:
Ok, with that bit of information I fully agree, that #272 is related.

Yet, I would vote against adding a respawn flag to the launchfiles as this in my opinion sends out the wrong signal to users that this driver might be rather unstable, which I think it is not.

For the debugging purposes you mentioned, it would probably have been sufficient to simply start the dashboard  client node separately.

steinmn:
Fair enough, I've removed the respawn arguments.

Where is the documentation on starting the dashboard client separately? This was news to me.

fmauch:
> Fair enough, I've removed the respawn arguments.
> 
> Where is the documentation on starting the dashboard client separately? This was news to me.

It's at least documented in the [ROS interface](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/doc/ROS_INTERFACE.md#dashboard_client) documentation.

fmauch (approved):
Fine with me

PR #453, opened 2021-08-13
Fix compile erros when using RoboStack (conda)

nberliner:
The RoboStack maintainers helped to fix compile errors when building the UR drivers using a conda environment. conda is, in my humble opinion, very useful and the suggested changes might be helpful for others.

For more information, please see RoboStack/ros-noetic#155 and RoboStack/ros-noetic#157 .

### DIFF
--- a/ur_calibration/CMakeLists.txt
+++ b/ur_calibration/CMakeLists.txt
@@ -11,6 +11,7 @@ find_package(catkin REQUIRED COMPONENTS
 find_package(Eigen3 REQUIRED)
 find_package(yaml-cpp REQUIRED)
 find_package(ur_client_library REQUIRED)
+find_package(console_bridge REQUIRED)
 
 # create alias for the '_INCLUDE_DIR' variable, to help Catkin export it
 set(YAML_CPP_INCLUDE_DIRS ${YAML_CPP_INCLUDE_DIR})
@@ -50,6 +51,7 @@ target_link_libraries(calibration_correction
   ${catkin_LIBRARIES}
   ${YAML_CPP_LIBRARIES}
   ur_client_library::urcl
+  console_bridge::console_bridge
 )
 
 install(DIRECTORY launch
--- a/ur_calibration/CMakeLists.txt
+++ b/ur_calibration/CMakeLists.txt
@@ -4,6 +4,8 @@ project(ur_calibration)
 ## Compile as C++11, supported in ROS Kinetic and newer
 add_compile_options(-std=c++11)
 
+add_definitions(-D__STDC_FORMAT_MACROS)
+
 find_package(catkin REQUIRED COMPONENTS
   roscpp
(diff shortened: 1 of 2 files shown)


fmauch:
This seems unusual at first.

The console_bridge issues should probably be gone with https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/74 merged.

As for the `-D__STDC_FORMAT_MACROS` definition: Why is this necessary?

nberliner:
Hi, thanks for having a look. To be honest, I don't know what `-D__STDC_FORMAT_MACROS` does or why it is necessary. It was suggested to me by the maintainer of RoboStack [here](https://github.com/RoboStack/ros-noetic/issues/155#issuecomment-897084877).

Maybe you could chime in on RoboStack/ros-noetic#157 to check where this fits best?

gavanderhoorn:
@fmauch wrote:
> As for the `-D__STDC_FORMAT_MACROS` definition: Why is this necessary?

From [How to printf uint64_t? Fails with: “spurious trailing ‘%’ in format”](https://stackoverflow.com/questions/8132399):

> The ISO C99 standard specifies that these macros must only be defined if explicitly requested.
> 
> ```c
> #define __STDC_FORMAT_MACROS
> #include <inttypes.h>
> 
> ... now PRIu64 will work
> 
> ```

fmauch:
Thanks @gavanderhoorn 

From my point of view none of the changes here should then be necessary with the latest versions.

@nberliner could you try building the driver without the changes with the latest library version (master or boost branch) and the latest master of this driver?

See the [README](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/master/README.md#alternative-all-source-build) about building the library from source, as well.

nberliner:
I've tried to build the driver as well as `ur_client_library` from source. Unfortunately, building the `ur_client_library` within the RoboStack conda environment seems to be rather complicated (see [here](https://github.com/RoboStack/ros-noetic/blob/master/patch/ros-noetic-ur-client-library.patch)) and `catkin_make_isolated` failed. I probably need to wait until the RoboStack maintainers include the updates.

Please feel free to close this PR, it's definitely beyond my expertise.

Thanks for looking into it :)

fmauch:
I should have a look at this again, before closing this.

fmauch (commented):
Sorry for not having answered for so long. I am still unsure whether I would like to add something that "only" makes it build on an unsupported build system.

The  `D__STDC_FORMAT_MACROS` thing seems to be something that should be fixed inside the conda builds as suggested in https://github.com/RoboStack/ros-noetic/issues/157.

If we could change the console_bridge dependency to a ordinary catkin dependency to `rosconsole` and not explicitly link against that but let `${catkin_LIBRARIES}` handle this, that would be great. Unfortunately, this is out of my scope to test.

PR #462, opened 2021-09-07
Prepare release v2.0.0

fmauch:
Prepare release with proper changelog and version numbers


fmauch:
Merging as it is only melodic testing failing.

PR #466, opened 2021-09-16
Create stale.yml

fmauch:
Mark stale issues and PRs automatically

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,29 @@
+# This workflow warns issues and PRs that have had no activity for a specified amount of time.
+#
+# You can adjust the behavior by modifying this file.
+# For more information, see:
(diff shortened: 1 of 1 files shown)


fmauch:
I've added automatic issue closing.

A couple of [exceptions](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/94efc8112b8191e08ec6cbe6de163c74ed3b7f37/.github/workflows/stale.yml#L29) should prevent handling issues with special labels.

@stefanscherzinger could you have a look whether you think this list is complete.

PRs are not closed automatically at all, but only tagged as "Stale".

stefanscherzinger:
@fmauch Yes, I'll look into it as soon as possible.

RobertWilbrandt (approved):
This seems reasonable to me, especially choosing relatively long stale and close times.

Issue #390, opened 2021-05-27
robot slowly moving without commands when pose_based_cartesian_traj_controller is running


A13X15:
# Summary
When using the controller pose_based_cartesian_traj_controller from the beta-testing branch, the robot is not holding its position when there are no commands, but moving slowly over time. May be a bug in the beta-software.

*Introduction to the issue*

# Versions
 - ROS Driver version: beta-testing, version from May 4th 2021
 - Affected Robot Software Version(s): 3.12
 - Affected Robot Hardware Version(s): UR10
 - Robot Serial Number:
 - UR+ product(s) installed: -
 - URCaps Software version(s): 1.0.4

# Impact
Robot may crash over time, if left unmonitored.

# Issue details

## Use Case and Setup
Tested the beta-branch with a real robot as cartesian control is a neat feature to have. Encoutered what I think is a bug with the 
pose_based_cartesian_traj_controller.

## Project status at point of discovered
 - followed the instructions to install the beta-testing branch 
 - read issue #373 to know how to use the new features
 - modified cartesian_trajectory_action_client.py to test the cartesian movement without random poses...
 - switched controller from scaled_pos_joint_traj_controller to pose_based_cartesian_traj_controller
 - test runs fine, cartesian movement is possible
 - however, robot moves after the script is finished (very slowly, but after a few minutes it becomes obvious)
 - only happen, when pose_based_cartesian_traj_controller is running; not happening without controller or with scaled_pos_joint_traj_controller running

## Steps to Reproduce
- use real hardware for testing (not sure how the simulation behaves, I used real hardware)
- start the pose_based_cartesian_traj_controller
- observe, what the robot is doing over 5-10 minutes

## Expected Behavior
- robot is not moving when there are no commands with pose_based_cartesian_traj_controller running

## Actual Behavior
- robot is slowly moving when there are no commands with pose_based_cartesian_traj_controller running

## Workaround Suggestion
- Idk, if it is a bug, fix it

A13X15:
Addition: When watching the teach pendant, I can also see the coordinates and angles changing slowly while the pose_based_cartesian_traj_controller is running.

It would be nice if someone can also test this to confirm, to make sure it is not an issue with my setup, but with the beta-software

fmauch:
@stefanscherzinger assigning this to you is OK?

stefanscherzinger:
@fmauch That's alright. I'll see how to get hands on a real robot in the next days.

[...]

PR #474, opened 2021-09-24
Added test for joint based cartesian trajectory interface

urmahp:
Added test of the robot moving slightly over time for both cartesian controllers.

This will test the issue #390, however it will fail on melodic, because the issue still exist on melodic.


fmauch:
keep open

fmauch:
@fmauch should review

PR #513, opened 2022-03-10
Migrate doc structure towards sphinx documentation

fmauch:
This gets built by the buildfarm automatically, which should provide a
a better entrypoint for users.

The generated documentation will be hosted on docs.ros.org

## ToDos:
 - [ ] Check for and fix broken links. Many of the moved files have been referenced from elsewhere.


fmauch:
yes, this should still be used!

fmauch:
still relevant

Issue #66, opened 2019-12-16
Feature: introduce a "roswtf" or "ros doctor"


gavanderhoorn:
Idea is to introduce a small utility that would be able to check whether configuration and setup of the robot controller and driver is as it should be, and prints a report of identified potential issues if this is not the case.

Compare this to `roswtf` and the new ROS 2 `doctor` commands.

List of things to check (@fmauch feel free to add to the list):

 - 'reverse port' configured in the *External Control*  urcap on the controller
 - IP configured in the *External Control*  urcap on the controller
 - whether URCap is running (in situations where it should be running)
 - system software version of the controller

Additional items could be added based on issues opened here on the tracker.

Intent is to help users avoid common configuration mistakes and by achieving this to reduce the number of support requests logged on the tracker.

fmauch:
related: https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/issues/147

urfeex:
Since ROS (1) is end of life this will not be worked on anymore.

PR #516, opened 2022-03-25
Add a network debugging tool to help with common connection problems

fmauch:
This aims at solving #66 

As questions about setting up the network configuration often popup and the approach to debug them is always the same, I've wrapped some of those steps into a helper script that should get started alongside with the driver.

## ToDo:
 - [ ] Check  the robot's installation for remote IP and port (I currently do now know how to do that)
 - [x] Add hints / guide on checking most common firewalls
 - [ ] Include into launchfiles with a flag

## Runtime checks
We could add runtime checks, as well, but that should probably go to a separate script / be activated through a service call
 - [ ] Check if controllers are running
 - [ ] Check if the robot can currently accept commands

### DIFF
--- a/ur_robot_driver/doc/check_firewalls.md
+++ b/ur_robot_driver/doc/check_firewalls.md
@@ -1,0 +1,96 @@
+# Check common firewalls on Linux
+
+When using this driver with a robot it is important that the robot can establish a connection to the
+machine running this driver on a couple of ports. In case there is a firewall active without a
+special configuration it will most likely block this connection.
+
+In order for this driver to work correctly, the firewall has to allow connections to the ports
+**50001, 50002 and 50003**. Obviously, as soon as you change the port arrangement manually, any firewall rules
+have to be changed accordingly.
+
+**NOTE: The steps shown in this tutorial will get your machine running with a robot connecting to
+it. It is not a full guide to Linux firewalls. Changing your firewall settings is potentially
+opening security holes in your local machine. In case of doubt, ask your local system administrator
+for advice before altering your firewall settings.**
+
+Changing your firewall settings will most likely require root (sudo) access to your local machine.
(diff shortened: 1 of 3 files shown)


gavanderhoorn:
High-level initial feedback: don't use `rospy` here.

I'd make this a stand-alone tool, which still checks the same things, but doesn't use ROS parameters or logging.

That decouples multiple problem domains, and would also make it possible to run this on OS which ROS doesn't OotB support / has different level of support for (like Windows).

An `argparse`-based implementation should easily allow this.

fmauch:
> High-level initial feedback: don't use `rospy` here.
> 
> I'd make this a stand-alone tool, which still checks the same things, but doesn't use ROS parameters or logging.
> 
> That decouples multiple problem domains, and would also make it possible to run this on OS which ROS doesn't OotB support / has different level of support for (like Windows).
> 
> An `argparse`-based implementation should easily allow this.

Actually I started with exactly that. Then I realized, that it might be good to query as much configuration as possible from the driver's setup (such as robot_ip, reverse_port, etc.) I am still unsure whether the benefits are worth it tying it to ROS, though. (Especially, since we want to use the same thing for ROS2, as well...)

gavanderhoorn:
There are other ways of getting that information, such as parsing the `.launch` files or loading the `.yaml` files.

Whether that's worth the extra effort I don't know.

There is certainly value in using the same parameters while attempting to diagnose problems, as one potential cause of those problems could exactly be those parameters.

fmauch:
I think I'll stick with getting the parameters from the parameter server, but I'll reduce the overall dependency on ROS so we can more easily re-use this for ROS2.

fmauch:
this is still on my whishlist...

urfeex:
This will no longer be pursued

PR #521, opened 2022-04-15
Minor update to display correct topic without _

williamnavaraj:


fmauch (approved):
Thanks @williamnavaraj

PR #535, opened 2022-05-24
Wait for controller action server in test_move.

adamheins:
Without waiting for the server to be ready, it is possible that the call to `trajectory_client.send_goal` happens before the server is ready to accept it. In my experience this doesn't happen in `test_move` script because the call to `self.ask_confirmation` takes enough time for the server to be ready. However, when I went to write my own motion script based on this one without the `self.ask_confirmation`, the server was not ready in time. It took sometime for me to debug, so seems like it may be useful to include for others.


fmauch (approved):
LGTM, thanks for adding this!

PR #544, opened 2022-06-10
Add optional topic rename for speed scaling factor

teapfw:
## What?
I've added the ability to change the default topic on which the speed scaling factor is published.
## Why?
This homogenizes the behavior, considering the wrench topic can currently be renamed. I've been working on a project that involves two UR arms running simultaneously, and it has been useful to maintain separate speed sliders rather than being forced to receive only one robot's slider state. In particular, if the robot which is not publishing state protective stops, I have difficulty stopping its related hardware because I have no programmatic indication of the fault.
## How?
Use a ROS Param just like the wrench topic is defined.
## Testing?
I have been using this in my project without any visible side effects.

### DIFF
--- a/ur_robot_driver/doc/ROS_INTERFACE.md
+++ b/ur_robot_driver/doc/ROS_INTERFACE.md
@@ -65,6 +65,10 @@ tf_prefix used for the robot.
 
 Parameter to set the id of the wrench frame, required if using multiple robots
 
+##### speed_scaling_id (default: "speed_scaling_factor")
+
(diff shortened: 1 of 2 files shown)


fmauch (changes requested):
Seems reasonable, thank you for the contribution!

Could you please add the parameter to the [ROS_INTERFACE](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/master/ur_robot_driver/doc/ROS_INTERFACE.md) documentation?

fmauch (approved):
LGTM, thank you!

PR #546, opened 2022-06-13
Implemented ROS topic to enable the freedrive mode.

t-schnell:
Implements a basic way to enable freedrive_mode via a ROS topic.

To test: send a message to the "enable_freedrive_mode" topic with a float value specifying seconds to be in freedrive mode.

Using "0.0" will enter freedrive_mode until the robot stops being moved continously (and at least 3 seconds to allow for start of the movement).

This is not a final implementation of this feature.

### DIFF
--- a/ur_robot_driver/include/ur_robot_driver/hardware_interface.h
+++ b/ur_robot_driver/include/ur_robot_driver/hardware_interface.h
@@ -213,6 +213,7 @@ protected:
   bool resendRobotProgram(std_srvs::TriggerRequest& req, std_srvs::TriggerResponse& res);
   bool zeroFTSensor(std_srvs::TriggerRequest& req, std_srvs::TriggerResponse& res);
   void commandCallback(const std_msgs::StringConstPtr& msg);
+  void freedriveCallback(const std_msgs::Float64Ptr& msg);
 
   std::unique_ptr<urcl::UrDriver> ur_driver_;
   std::unique_ptr<DashboardClientROS> dashboard_client_;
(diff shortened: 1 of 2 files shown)


fmessmer:
Thanks for the feature draft!
We'll definitely try this - once our HW is free for testing

What's the motivation for topic activation - over e.g. service/action?
I think an action interface would be nice as the driver/server could deny to switch to freedrive mode or abort in case of error...also the client could cancel freedrive mode early if no longer needed...and the duration could still be set in the goal resulting in a succeeded case to achieve the same behavior as with the plain topic-based interface...

fmessmer:
Finally, I was able to test this feature on our HW and it worked out of the box :+1: 
So far, we don't have any explicit requirements for what the interface (topic, service, action) should be - as we are still experimenting...
Thus, I'd be ok to get this merged and modify/extend the interface lateron....

fmessmer (approved):

fmauch:
Before merging, I think at least some basic documentation should be added.

fmessmer:
@ github-actions
I'm still interested in this feature!
(hope this keeps it open :wink:)

fmauch:
This is currently being worked on in https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/138 and is not forgotten ;-)

fmessmer:
this github-actions is really annoying!

ZdenekM:
The feature would be great; I'm only voting for an action interface, which makes more sense to me.

fmessmer:
we recently proposed https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/718

Issue #349, opened 2021-04-03
Compile error building beta-testing branch


robotrapta:
# Summary

I can't get the beta-testing branch to build following the README instructions.  My C++ is super rusty (no pun intended) but it seems like the ur_robot_driver package isn't finding the headers from Universal_Robots_Client_Library or something.  The errors are like `urcl::comm::TrajectoryControlMessage has not been declared` even though it's clearly there in the client library code, which is building properly.

# Versions
 - ROS Driver version: tested on both melodic and noetic - identical results
 - Affected Robot Software Version(s): n/a
 - Affected Robot Hardware Version(s): n/a
 - Robot Serial Number: n/a 
 - UR+ product(s) installed: n/a 
 - URCaps Software version(s): n/a

# Impact
Blocked.  Can't use beta-testing branch at all.

## Steps to Reproduce
Following the readme instructions:

```
$ source /opt/ros/melodic/setup.bash
$ mkdir -p catkin_ws/src && cd catkin_ws
$ git clone -b beta-testing-boost https://github.com/UniversalRobots/Universal_Robots_Client_Library.git src/Universal_Robots_Client_Library
$ git clone -b beta-testing https://github.com/UniversalRobots/Universal_Robots_ROS_Driver.git src/Universal_Robots_ROS_Driver
$ git clone -b beta-testing https://github.com/fzi-forschungszentrum-informatik/cartesian_ros_control.git src/cartesian_ros_control
$ git clone -b calibration_devel https://github.com/fmauch/universal_robot.git src/fmauch_universal_robot
$ sudo apt update -qq
$ rosdep update
$ rosdep install --from-paths src --ignore-src -y
$ catkin_make_isolated
```

gives build error messages:

```
/home/astro/beta_ws/src/Universal_Robots_ROS_Driver/ur_robot_driver/src/hardware_interface.cpp:675:19: error: ‘class urcl::UrDriver’ has no member named ‘writeTrajectoryControlMessage’
       ur_driver_->writeTrajectoryControlMessage(urcl::comm::TrajectoryControlMessage::TRAJECTORY_NOOP);
                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/astro/beta_ws/src/Universal_Robots_ROS_Driver/ur_robot_driver/src/hardware_interface.cpp: In member function ‘virtual void ur_driver::HardwareInterface::write(const ros::Time&, const ros::Duration&)’:
[...]

PR #549, opened 2022-06-21
Add optional calibration step for all-source build

captain-yoshi:
With the [all-source build](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver#alternative-all-source-build), the [proposed calibration launch pkg](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/tree/master/ur_calibration#creating-a-calibration--launch-package-for-all-local-robots) fails to build, because it searches for the `ur_client_library` in the binary global space `/opt/ros/noetic`. With the proposed changes, it searches the catkin workspace beforehand and successfully finds the library. Using ROS-Noetic with catkin-tools.
```bash
CMake Error at /opt/ros/noetic/share/catkin/cmake/catkinConfig.cmake:83 (find_package):
  Could not find a package configuration file provided by "ur_client_library"
  with any of the following names:

    ur_client_libraryConfig.cmake
    ur_client_library-config.cmake

  Add the installation prefix of "ur_client_library" to CMAKE_PREFIX_PATH or
  set "ur_client_library_DIR" to a directory containing one of the above
  files.  If "ur_client_library" provides a separate development package or
  SDK, be sure it has been installed.
Call Stack (most recent call first):
  CMakeLists.txt:10 (find_package)
```


fmauch:
I never had problems building it without the modification, but the supposed change might make sense to actually merge into the CMakeLists.txt directly...

fmauch:
Oh, wait, this should solve #349? That would be aweseome!

captain-yoshi:
I don't think so. The `CMakeLists.txt` seems to bet set properly in both `ur_calibration` and `ur_robot_driver` (even from 2 years ago). On my end and without the proposed example_organization_ur_launch package it runs fine. Maybe a complete clean of the workspace would have helped out.

The problem I have is only when creating the `example_organization_ur_launch` because of this line
``` cmake
$ catkin_create_pkg example_organization_ur_launch ur_client_library \
-D "Package containing calibrations and launch files for our UR robots."
```
which adds the **non catkin package** `ur_client_library` as a catkin component
``` cmake
find_package(catkin REQUIRED COMPONENTS
  ur_client_library
)
```
which overshadows the local workspace build of `example_organization_ur_launch`.

Will update a better fix soon.

fmauch:
Ah, now I get it. In fact, the dependency on ur_client_library is indeed wrong. It should be `ur_robot_driver` instead. Then, it also will not cause the problem you are mentioning.

captain-yoshi:
@fmauch Changed the wrong dependency and it works as expected.

fmauch (approved):
Looking good. Thanks @captain-yoshi for sticking with this.

PR #552, opened 2022-06-28
test_move: Load controller only if it is not already loaded

fmauch:
Otherwise we get an ugly error message that the target controller is already loaded.


RobertWilbrandt (approved):
This seems reasonable

PR #554, opened 2022-06-29
WIP: Use anchors and aliases for each joint in controller config + cleanup

captain-yoshi:
This PR uses anchors and aliases on each robot joint and sets indentation level to 2 spaces (there were mixed indentation). This makes the following benefits:
- Same indentation level (set to 2 but can be anything)
- Consistency with string values (removed dquotes)
- PID and trajectory/goal section is easier on the eyes...
- Adding a `prefix` to joint names is easier (less search and replace)

Only did the changes for the UR3. If you think this change could be merged, I will modify the other UR config files.

### DIFF
--- a/ur_robot_driver/config/ur3_controllers.yaml
+++ b/ur_robot_driver/config/ur3_controllers.yaml
@@ -1,142 +1,142 @@
 # Settings for ros_control control loop
 hardware_control_loop:
-   loop_hz: &loop_hz 125
+  loop_hz: &loop_hz 125
 
 # Settings for ros_control hardware interface
 ur_hardware_interface:
-   joints: &robot_joints
-     - shoulder_pan_joint
-     - shoulder_lift_joint
-     - elbow_joint
-     - wrist_1_joint
-     - wrist_2_joint
-     - wrist_3_joint
+  joints: &robot_joints
+    - shoulder_pan_joint
+    - shoulder_lift_joint
+    - elbow_joint
+    - wrist_1_joint
+    - wrist_2_joint
+    - wrist_3_joint
 
 # Publish all joint states ----------------------------------
 joint_state_controller:
-   type:         joint_state_controller/JointStateController
-   publish_rate: *loop_hz
+  type:         joint_state_controller/JointStateController
+  publish_rate: *loop_hz
 
 # Publish wrench ----------------------------------
 force_torque_sensor_controller:
-   type:         force_torque_sensor_controller/ForceTorqueSensorController
-   publish_rate: *loop_hz
+  type:         force_torque_sensor_controller/ForceTorqueSensorController
+  publish_rate: *loop_hz
 
 # Publish speed_scaling factor
 speed_scaling_state_controller:
-   type:         scaled_controllers/SpeedScalingStateController
-   publish_rate: *loop_hz
+  type:         scaled_controllers/SpeedScalingStateController
+  publish_rate: *loop_hz
 
 # Joint Trajectory Controller - position based -------------------------------
 # For detailed explanations of parameter see http://wiki.ros.org/joint_trajectory_controller
 scaled_pos_joint_traj_controller:
-   type: position_controllers/ScaledJointTrajectoryController
-   joints: *robot_joints
-   constraints:
-      goal_time: 0.6
-      stopped_velocity_tolerance: 0.05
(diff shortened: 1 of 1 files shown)


ur_robot_driver/config/ur3_controllers.yaml
```diff
@@ -1,166 +1,166 @@
...
+  joints: *robot_joints
+  constraints:
+    goal_time: 0.6
+    stopped_velocity_tolerance: 0.05
+    *joint1: {trajectory: 0.2, goal: 0.1}
+    *joint2: {trajectory: 0.2, goal: 0.1}
+    *joint3: {trajectory: 0.2, goal: 0.1}
+    *joint4: {trajectory: 0.2, goal: 0.1}
+    *joint5: {trajectory: 0.2, goal: 0.1}
+    *joint6: {trajectory: 0.2, goal: 0.1}
```
fmauch, line 43: I am not sure whether using anchors for this really improves things. Yes, it is less code duplication, but in my opinion this is also less readable.

fmauch (commented):
Thanks for taking the effort diving into this. Please see my comments and if this is going to be merged, all controller files should be handled I think.

gavanderhoorn:
Something to keep in mind: anchors et al. are not part of the YAML standard.

They never made it past the draft/proposed state of YAML 1.1 (IIRC).

ur_robot_driver/config/ur3_controllers.yaml
captain-yoshi: If someone uses the prefix below in their urdf:

``` xml
<xacro:ur3_robot prefix="ur3_"/>
```

All the joints must be renamed, with anchors/aliases you only need to rename 6 joint names. Would something like this be better ?
``` yaml
    *shoulder_pan_joint:  {trajectory: 0.2, goal: 0.1}
    *shoulder_lift_joint: {trajectory: 0.2, goal: 0.1}
    *elbow_joint:         {trajectory: 0.2, goal: 0.1}
    *wrist_1_joint:       {trajectory: 0.2, goal: 0.1}
    *wrist_2_joint:       {trajectory: 0.2, goal: 0.1}
    *wrist_3_joint:       {trajectory: 0.2, goal: 0.1}
```

gavanderhoorn: have you considered using [subst_value](http://wiki.ros.org/roslaunch/XML/rosparam#Attributes)? I don't believe you'd need to manually rename anything.

captain-yoshi: This would work with the `/my_controller/joints` because they are values, but I would need to remap the `constraints`, `gains` and `velocity_ff` namespaces:
``` bash
 * /pos_joint_traj_controller/constraints/ur3_elbow_joint/goal: 0.1
 * /pos_joint_traj_controller/constraints/ur3_elbow_joint/trajectory: 0.2
 * /pos_joint_traj_controller/constraints/ur3_shoulder_lift_joint/goal: 0.1
 * /pos_joint_traj_controller/constraints/ur3_shoulder_lift_joint/trajectory: 0.2
 * /pos_joint_traj_controller/constraints/ur3_shoulder_pan_joint/goal: 0.1
 * /pos_joint_traj_controller/constraints/ur3_shoulder_pan_joint/trajectory: 0.2
 * /pos_joint_traj_controller/constraints/ur3_wrist_1_joint/goal: 0.1
 * /pos_joint_traj_controller/constraints/ur3_wrist_1_joint/trajectory: 0.2
 * /pos_joint_traj_controller/constraints/ur3_wrist_2_joint/goal: 0.1
 * /pos_joint_traj_controller/constraints/ur3_wrist_2_joint/trajectory: 0.2
 * /pos_joint_traj_controller/constraints/ur3_wrist_3_joint/goal: 0.1
 * /pos_joint_traj_controller/constraints/ur3_wrist_3_joint/trajectory: 0.2
```
When adding new controllers I would have to remap the new controller namespace. With the anchors solution I don't have to bother, it scales well when adding new controllers. I may be missing something...

gavanderhoorn: I don't understand what you write here.

But perhaps I also don't fully understand what you're trying to achieve.

fmauch:
I don't necessarily want to have this closed by a bot, hence commenting on this.

Issue #76, opened 2020-01-10
Add missing calls from dashboard server


fmauch:
There are a couple of calls missing from the dashboard server, especially since Versions 3.12/5.6

 - [ ] PolyscopeVersion
 - [ ] setUserRole <role>
 - [ ] get serial number
 - [ ] get robot model
 - [ ] get operational mode
 - [ ] is in remote control
 - [ ] set operational mode <mode>
 - [ ] clear operational mode

MinhDangTG:
Hello fmauch,

it seems liked an already-answered question, but do we have a service/method to turn on the electricity of the control box instead of pressing the power button on the teach pendant? since we had the shutting down service, I though that it would be handy if we had a similar method of the other way around. 

It seems almost impossible since the shutting-down will lose all the connection between the robot and ROS. A solution may be connecting another button or an arduino to pins in the control box (liked the safeguard stop) to booting the control box. 

Thanks.

fmauch:
> Hello fmauch,
> 
> it seems liked an already-answered question, but do we have a service/method to turn on the electricity of the control box instead of pressing the power button on the teach pendant? since we had the shutting down service, I though that it would be handy if we had a similar method of the other way around.
> 
> It seems almost impossible since the shutting-down will lose all the connection between the robot and ROS. A solution may be connecting another button or an arduino to pins in the control box (liked the safeguard stop) to booting the control box.
> 
> Thanks.

I think, this is not possible to do. 

In the scope of this issue it is absolutely impossible, as obviously if the robot is switched off, there is no dashboard server to send a command to.

A physical wiring as you suggested it might work, if you can get access to the necessary pins inside the controller box. I guess you can, as I have seen UR arms integrated on mobile robot platforms, but there was still another physical button to press.

Another solution might be Wake-On-LAN, but I am not sure whether this is supported by the controller PC / would require modification of the controller PC.

Please let's not discuss this further inside this issue, as this is out of scope for the dashboard server.

urfeex:
Since ROS (1) is end of life this will not be worked on anymore. See Add missing calls from dashboard server UniversalRobots/Universal_Robots_ROS2_Driver#1649 for ROS 2.

PR #561, opened 2022-07-28
Dashboard service to query whether the robot is in remote control

mgkwon:
This PR adds one of the missing calls, `is in remote control`, from the dashboard server.

This is accessible through `raw request` but I thought it would be better to have it here. I am currently using it to alert the remote operator. Sometimes people forget to set it back to remote mode after the manual operation.

After preparing the pr, I found out there is an issue-#76 raised to implement additional missing calls. @fmauch If you see it reasonable to add the remaining calls, I can prepare separate PRs for them.

### DIFF
--- a/ur_robot_driver/doc/ROS_INTERFACE.md
+++ b/ur_robot_driver/doc/ROS_INTERFACE.md
@@ -736,6 +736,10 @@ Query whether there is currently a program running
 
 Query whether the current program is saved
 
+##### dashboard/in_remote_control ([ur_dashboard_msgs/IsInRemoteControl](http://docs.ros.org/api/ur_dashboard_msgs/html/srv/IsInRemoteControl.html))
+
+Query whether the robot is in remote control
+
 ##### dashboard/program_state ([ur_dashboard_msgs/GetProgramState](http://docs.ros.org/api/ur_dashboard_msgs/html/srv/GetProgramState.html))
 
 Service to query the current program state
--- a/ur_robot_driver/doc/ROS_INTERFACE.md
+++ b/ur_robot_driver/doc/ROS_INTERFACE.md
@@ -736,7 +736,7 @@ Query whether there is currently a program running
 
 Query whether the current program is saved
 
-##### dashboard/in_remote_control ([ur_dashboard_msgs/IsInRemoteControl](http://docs.ros.org/api/ur_dashboard_msgs/html/srv/IsInRemoteControl.html))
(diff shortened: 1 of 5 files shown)


fmauch:
Thank you for the contribution! Let me review this PR and then you are more than welcome to add an additional PR containing more from #76 :-)

ur_robot_driver/src/dashboard_client_ros.cpp
```diff
@@ -120,6 +120,10 @@ DashboardClientROS::DashboardClientROS(const ros::NodeHandle& nh, const std::str
   // Query whether the current program is saved
   is_program_saved_service_ = nh_.advertiseService("program_saved", &DashboardClientROS::handleSavedQuery, this);
 
+  // Query whether the robot is in remote control
+  is_in_remote_control_service_ =
+      nh_.advertiseService("in_remote_control", &DashboardClientROS::handleRemoteControlQuery, this);
```
fmauch, line 125:
```suggestion
      nh_.advertiseService("is_in_remote_control", &DashboardClientROS::handleRemoteControlQuery, this);
```

I'd prefer if the service names reflect the dashboard commands as close as possible. I know that we haven't been on spot there in the past and also the commands themselves are not consistent in terms of syntax, but it would be great to have a look on this in the future.

fmauch (changes requested):
In general this looks very good, thank you! I've added a couple of (minor) suggestions. With them applied, I think this is ready to merge.

mgkwon:
@fmauch Thanks for the suggestions. I will later prepare an MR with the rest of the calls with your suggestions regarding names and e-series specifics in mind :)

ur_robot_driver/doc/ROS_INTERFACE.md
```diff
@@ -736,6 +736,10 @@ Query whether there is currently a program running
 
 Query whether the current program is saved
 
+##### dashboard/in_remote_control ([ur_dashboard_msgs/IsInRemoteControl](http://docs.ros.org/api/ur_dashboard_msgs/html/srv/IsInRemoteControl.html))
```
fmauch, line 739:
```suggestion
##### dashboard/is_in_remote_control ([ur_dashboard_msgs/IsInRemoteControl](http://docs.ros.org/api/ur_dashboard_msgs/html/srv/IsInRemoteControl.html))
```

fmauch (commented):
looks good so far, I'll have to do a manual test before merging, though. As I am currently on vacation, this will probably not happen too soon.

fmauch (approved):
I've just tested this manually and adapted the documented service name. Looks good to me!

PR #566, opened 2022-08-12
Updated transformForceTorque to handle wheter it is a cb3 or an e-Ser…

urmahp:
…ies robot

The force torque is returned at the tool flange on e-series robots and at the tcp for CB3, this is now handled correctly, so that all force/torque measurements will be relative to the active TCP

The function tcp_offset isn't available before software version 3.12.0 and 5.5.1, meaning that the driver wont work with older software versions, but it is needed to transform the force torque measurements correctly. 

The tests will fail, since this PR requires changes from this [PR](https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/110) in the client library.


fmauch:
I'll give this some lab testing before approving, but it is looking good to me!

fmauch (approved):
Tested on a UR3e and UR10 where it produced sane output. Thanks @urmahp

Issue #562, opened 2022-08-01
Setting tool voltage at runtime is not supported


fmauch:
The [`setIO`](https://docs.ros.org/en/noetic/api/ur_msgs/html/srv/SetIO.html) service supports setting the tool voltage. However, this is currently [not supported](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/1eb7bf94a163a885d7bd49040f62f4160ce06fb9/ur_robot_driver/src/hardware_interface.cpp#L1100) by our wrapper.

As setting this is not available through RTDE we would have to either
 - Add that to our custom communication protocol
 - Use script code to control that. Drawback of this method would be that on e-Series this would only work when the robot is in remote_control mode.

urfeex:
As ROS 1 is end of life, this is not getting to be implemented. ROS 2 does support setting tool voltage through that service.

PR #567, opened 2022-08-16
Updated set payload, zero ftsensor and set tool voltage to use the ur…

urmahp:
…driver

This makes it possible to call the commands when the robot is in local control if the external control script is running on the robot.

This is a solution to #562. 

This PR requires the changes from this [PR](https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/111) in the client library


fmauch (approved):
Tested this and it seems to work nicely. Waiting for melodic and noetic sync before merging.

ur_robot_driver/launch/ur_common.launch
```diff
@@ -41,6 +42,7 @@
     <arg name="reverse_port" value="$(arg reverse_port)"/>
     <arg name="script_sender_port" value="$(arg script_sender_port)"/>
     <arg name="trajectory_port" value="$(arg trajectory_port)"/>
+    <param name="script_command_port" value="$(arg script_command_port)"/>
```
fmauch, line 45: Missed that one earlier, stumbled upon this in #582
```suggestion
    <arg name="script_command_port" value="$(arg script_command_port)"/>
```

fmauch:
I've cherry picked the docs update from #582, thanks @miguelprada

fmauch (approved):
With the recent changes we would be back to waiting for the sync again.

PR #574, opened 2022-10-03
Fix universal_robot clone destination in BUILDING

j3soon:
`universal_robot` should be cloned to `catkin_ws/src` instead of `catkin_ws`.

The current command clones it to `catkin_ws`, which results in the following error:

```
ERROR: the following packages/stacks could not have their rosdep keys resolved
to system dependencies:
ur_robot_driver: Cannot locate rosdep definition for [ur_description]
```


fmauch (approved):
Absolutely! This was missed in #560. Thanks for spotting this.

PR #576, opened 2022-10-05
Remove urcap version and add resources link in readme and docs

tony2guo:
Avoid editing docs every time when urcap version is changed.


fmauch:
Thank you for the suggestion. While rewriting this: It might be a good option to simply link to the URCap release instead of dragging it along with the driver. But that has to be discussed...

fmauch:
The external_control URCap comes from https://github.com/UniversalRobots/Universal_Robots_ExternalControl_URCap/releases. We've discussed that internally and decided to go ahead and remove the URCaps from the driver's package.

tony2guo:
Thanks!

PR #577, opened 2022-10-06
[ROS-O] Drop old C++ compiler flags

v4hn:
C++14 has been the default for clang and gcc since before Ubuntu 18.04 and enforcing the old standard breaks on systems with a current version of log4cxx (Ubuntu 22.04 +, Arch Linux, Lunar Linux, Debian Testing, ...)


fmauch (approved):
Since the lowest version we support is ROS melodic on ubuntu 18.04 with minumum g++ 7.4.0 which has c++14 as default, I'd be fine with making that change. Thank you for the suggestion.

PR #582, opened 2022-10-18
Script command port

miguelprada:
I suddenly started having issues in a dual-UR setup in our lab. After a bit of digging, I found that a new reverse connection from the robot to the PC was ["recently" added to the client library](https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/111). Since this new port is not configurable at the ROS driver level, it's currently impossible to run multi-UR ROS setups without all but one UR driver failing to initialize because of port collisions.

This PR adds this new parameter to the ROS driver node to allow configuring the port to use. It also uses the new `zeroFTSensor` method in `UrDriver`, which has the benefit of working both when Polyscope is in Local or Remote modes. I was not so sure of how the new `setToolVoltage` and `setPayload` in `UrDriver` could be used from the ROS driver, so I did nothing with those.

### DIFF
--- a/ur_robot_driver/doc/ROS_INTERFACE.md
+++ b/ur_robot_driver/doc/ROS_INTERFACE.md
@@ -49,6 +49,10 @@ Robot description launch file.
 
 IP address by which the robot can be reached.
 
+##### script_command_port (default: "50004")
+
+Port that will be opened to forward script commands to the robot when in local control mode.
+
 ##### script_sender_port (default: "50002")
 
 The driver will offer an interface to receive the program's URScript on this port. If the robot cannot connect to this port, `External Control` will stop immediately.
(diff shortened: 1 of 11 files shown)


miguelprada:
And obviously, there was already #567 to fix this. Sorry for not doing my homework :sweat_smile:

gavanderhoorn:
> And obviously, there was already #567 to fix this.

could be, but your PR is (subjectively) cleaner :)

miguelprada:
> could be, but your PR is (subjectively) cleaner :)

Appreciate your comment, but since the other one came first and packed more features, I reasoned it was logical to close this one. I'm glad that taking some time to make meaningful commits made it easy to cherry pick and combine both.

PR #585, opened 2022-11-10
Use the universal_robot melodic-devel branch

RobertWilbrandt:
With https://github.com/ros-industrial/universal_robot/pull/617 merged there is no more reason to use the staging branch.


fmauch (approved):
LGTM!

Technically, we could refer to the `melodic` branch in the not-all-source build, but that will be removed once we have synchronized binary releases, anyway.

PR #588, opened 2022-11-22
use released ur_description instead of building it from source

fmauch:
This PR aims at providing the new default stable branch. Since we released the universal_robots repository, we can start releasing the driver.

As a first step, I removed the rosinstall entries. At this stage the builds using the testing repo should succeed without any upstream workspace.


fmauch:
After discussing this with @RobertWilbrandt I'll close this. In our opinion it indeed does make sense to build this repo against the latest versions of support and library packages in the CI. We could add a separate binary-only build in future, as we did for the ROS2 driver.

PR #593, opened 2022-12-06
delete ros_control.urscript

fmauch:
We've been using the script from the library for a while now


RobertWilbrandt (approved):
Good catch!

PR #594, opened 2022-12-12
Add license file at root and in msgs package

fmauch:
As suggested in https://github.com/ros/rosdistro/pull/35593#discussion_r1044750667


RobertWilbrandt (approved):
LGTM

PR #596, opened 2022-12-16
set receive timeout after connecting the client

steffen-roperobotics:
The dashboard client connect function resets the receive timeout to 1 s. To avoid that the receive timeout is reset the setReceiveTimeout call is moved to after the client is connected.


fmauch:
Looks like you are right. Thanks for spotting this and providing the fix as a PR!

fmauch (approved):
LGTM. Since we [set the timeout](https://github.com/UniversalRobots/Universal_Robots_Client_Library/blob/09adab9aea47c6e660d9bf18db5f69765dd2fa4d/src/ur/dashboard_client.cpp#L82-L85) in the `connect()` method this has to be done afterwards.

fmauch:
Before merging I'd like to test this on a real hardware. I might have never switched installations using the dashboard client before. If a higher timeout is needed in order to switch installations, we should probably also change the default value.

fmauch:
Yes, this is still relevant. I should test and merge this!

fmauch:
I finally got around testing this on real hardware. Thank you @steffen-roperobotics for providing this!

PR #603, opened 2023-01-09
Added service to set force mode through ROS

urmarp:
Added service to set force mode through ROS


urfeex:
Closing this as a more up-tp-date version exists in #707

PR #609, opened 2023-01-23
Move controller_stopper to ur_robot_driver

fmauch:
Since a standalone package with that name was declined during the release process, we decided to move the controller_stopper package over to the driver.


RobertWilbrandt (approved):
After a few revisions this seems good now. Thanks!

PR #612, opened 2023-01-23
Prepare changelog for upcoming release

RobertWilbrandt:
Unfortunately 2.1.1 was still missing #594 which we want to include for the release.


fmauch (approved):
lgtm

PR #653, opened 2023-08-15
Add issue templates for bugs and features

urrsk:


RobertWilbrandt:
How does this interact with the `.github/issue_template.md` issue template we currently have for this repo? I would guess that it should get removed?

urrsk:
Yes we should remove that one. But lets do that after we have seen have this is used.

fmauch (approved):
Let's go with that for now.

PR #655, opened 2023-08-24
Update documentation

RobertWilbrandt:
Our documentation has gone a bit out of date. We should update it and restructure it to represent the current driver state (e.g. recommend binary installation etc.).

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,8 +88,8 @@ This repository contains the new **ur_robot_driver** and a couple of helper pack
   * **ur_robot_driver**: The actual driver package.
 
 ## Requirements
-This driver requires a system setup with ROS. It is recommended to use **Ubuntu 18.04 with ROS
-melodic**, however using Ubuntu 20.04 with ROS noetic should also work.
+This driver requires a system setup with ROS. It is recommended to use **Ubuntu 20.04 with ROS
+noetic**.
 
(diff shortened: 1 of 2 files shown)


fmauch:
I think one more thing to include in this PR would be adding the binary installation instructions on top of building from source, as this should be the default installation method.

README.md
```diff
@@ -85,38 +85,36 @@ This repository contains the new **ur_robot_driver** and a couple of helper pack
...
+noetic. If you want to use a UR robot arm with ROS 2, please see the
+[ROS 2 driver](https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/).
+2. Install the driver using
+```bash
+sudo apt install ros-${ROS_DISTRO}-ur-robot-driver
+```
+## Building from source
 
-## Building
+Before building from source make sure that you actually need to do that. We recommend you install the driver package from source unless you want to join development and submit changes.
```
fmauch, line 110:
```suggestion
Before building from source make sure that you actually need to do that. We recommend you install the driver from its binary packages unless you want to join development and submit changes.
```

fmauch (changes requested):
I think the warning in the from-source build seems to be at least confusing and needs to be changed.

fmauch (approved):
Thank you, that really adds some needed documentation.

PR #656, opened 2023-08-24
fix typo and use ros logging

simonschmeisser:
as discovered in https://github.com/ros-industrial/kuka_experimental/pull/257


fmauch (approved):
thanks for pointing that out.

PR #664, opened 2023-10-11
Bump required cmake version to fix the CMake warning about CMP0048

fmauch:
See https://github.com/ros/catkin/pull/1052 for details


RobertWilbrandt (approved):
Looks good to me, according to [REP 3](https://www.ros.org/reps/rep-0003.html#noetic-ninjemys-may-2020-may-2025) we should be able to assume at least CMake 3.13.4 on supported plattforms.

PR #667, opened 2023-10-26
Modified the language of README.md file.

rohan37kumar:
Better language helps in understanding and more productivity.

I'm happy to contribute in this project.


fmauch (changes requested):
Thank you for contributing to this and helping to improve one of our main entry points for new users. As I am not a native English speaker myself my judgement on this can only go so far (especially on the commas), but in some of the points I have risen I am pretty sure that they are actually wrong.

PR #685, opened 2024-02-20
Update robot SW requirements

urrsk:
Align the documented required Robot SW version to the same as the current version of the client library


fmauch (approved):
Thanks!

PR #686, opened 2024-02-22
Added service to enable and disable tool contact.

URJala:
Also added service to get robot software version.

This PR requires the changes from [PR](https://github.com/ros-industrial/ur_msgs/pull/24). 

This adds support for tool contact as proposed in #678.


urfeex:

PR #695, opened 2024-04-11
Implemented get version service

URJala:
And integrated it with the tool contact test. This service will obtain the software version of the robot, which can be useful for certain functionalities that only work on e-series robots, for example. This PR depends on [PR](https://github.com/ros-industrial/ur_msgs/pull/24) for functionality and [PR](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/686) for the integration test to work.


fmauch (changes requested):
This needs to be updated to the changed service definition in https://github.com/ros-industrial/ur_msgs/pull/24

fmauch (approved):
This is looking good. Waiting to merge until the message package is updated.

PR #707, opened 2024-07-09
Force mode arguments

URJala:
add damping and gain scale arguments to ROS service call. These can be set with each force mode call when [PR](https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/208) is implemented. 
This is also dependent on [PR](https://github.com/ros-industrial/ur_msgs/pull/27) for the service definition.


urfeex (commented):
This implementation has the same issues as #718: When this is used while the JTC is active, the robot will try to get back to the latest commanded configuration once force mode is disabled. We cannot merge it like this.

urfeex:

PR #713, opened 2024-07-18
Updated the UR family photo on the readme

urrsk:
Update the picture to also include the UR20 and UR30


fmauch (approved):
Thanks for providing this!

urrsk:
Maybe we should make it bigger..?

fmauch:
We definitively could do that. The centering approach that we have in the source code seems not to be working fine.

Issue #525, opened 2022-05-06
Feature idea: add freedrive_mode as interface


fmauch:
Adding an interface for activating the freedrive mode would probably be beneficial especially in situations where you don't have a teach pendant available.

As activating and deactivating freedrive_mode is available in ur_script, this could be easily added.

I suggest using a topic interface with a keepalive mechanism, so a user would have to regularly send a message to activate it. If input on that topic times out, it should be deactivated. The reason for that would be that it would have the same "deadman" mechanism as on the TP.

By providing a custom message and not only a simple bool topic we could restrict the freedrive functionality to only certain axes. It would probably be best to use URs `feature` notation for that.

It should be fairly easy to get a proof of concept running with this.

A few culprits to keep in mind:
 - This would only work if the robot is in remote_control mode
 - Activating this will stop script execution of other scripts, meaning users would have to restart external control afterwards. (Same goes for using the teach button on the TP, though)
 - The URScript function `teach_mode` has been migrated to `freedrive_mode`. We should check from which version on

@fmessmer, @floweisshardt FYI

floweisshardt:
+1 for enabling this feature through a ROS interface!

Thanks @fmauch for wrapping this up.

shuobh:
+1 This would be really useful!
I currently ran into an issue when switching between freedrive (with URScript) and remote control. The way I do so is:

- rosservice call stop -> wait 0.5s -> publish urscript to change to freedrive
- rosservice call stop -> wait 0.5s -> rosservice call play -> wait 0.5s -> start publishing trajectory using actionlib

If I do not have the 0.5s delay in between, although rosservice returns success in stop and play, it is not really ready to run new commands. 0s delay fails consistently and 0.2s delay fails randomly. So far 0.5 is the magic number.
If freedrive_mode is implemented, then I assume it wouldn't be necessary to add delays anymore.

However, the open question here is why UR is not immediately ready after the rosservice call? If there is a determined delay, why can't that be checked within the rosservice and only return when it is really ready to operate?

urfeex:
This has been implemented for ROS 2 by now. Closing this for ROS 1.

PR #573, opened 2022-09-28
Added freedrive mode functionality

MMT-Testbench:
'/enable_freedrive_mode' topic receives std_msgs/String:
- "[X,Y,Z,RX,RY,RZ]" where X,Y,Z,RX,RY,RZ = 1 or 0 for axis constraint

PR #718, opened 2024-07-31
Add free drive mode support

schmid-jn:
Suggested implementation of a free drive mode capability.
ref: https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/issues/525

### DIFF
--- a/ur_robot_driver/include/ur_robot_driver/hardware_interface.h
+++ b/ur_robot_driver/include/ur_robot_driver/hardware_interface.h
@@ -213,6 +213,7 @@ protected:
   bool resendRobotProgram(std_srvs::TriggerRequest& req, std_srvs::TriggerResponse& res);
   bool zeroFTSensor(std_srvs::TriggerRequest& req, std_srvs::TriggerResponse& res);
   void commandCallback(const std_msgs::StringConstPtr& msg);
+  void freeDriveModeCallback(const std_msgs::BoolConstPtr& msg);
   bool setPayload(ur_msgs::SetPayloadRequest& req, ur_msgs::SetPayloadResponse& res);
   bool activateSplineInterpolation(std_srvs::SetBoolRequest& req, std_srvs::SetBoolResponse& res);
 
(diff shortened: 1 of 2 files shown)


HannesBachter (approved):
tested on [AMC-H](https://www.4am-robotics.com/en/products/cobot-indoor/amc-h) :+1:

fmauch:
Hi, thank you for contributing to this project. However, there are already two PRs open with that functionality. If you have anything to add there, please comment in those.

schmid-jn:
@fmauch I could find https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/546 and https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/573. Both prs are open since two years and no one is working on it. Also this prs using a different approach enabling free drive via ur script and not with the new function in the client library.

fmessmer:
@fmauch 
please provide a strategy for timely support of adding the free-drive feature
we really would like to see this feature merged and released

the initial prs/drafts
- https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/546
- https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/573

already have been working for us - as commented - but were not merged due to https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/546#issuecomment-1405140173 and the work done in https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/138

now https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/pull/718 makes use of said URCL extension and still was closed in favor of the older, declined PRs above...

fmessmer:
@fmauch 
please help us understand why free-drive mode does not get supported by the official repo - no matter what concept is provided?

fmauch:
> @fmauch please help us understand why free-drive mode does not get supported by the official repo - no matter what concept is provided?

One problem is time. We are currently following up on this and plan to get it to a mergeable state until ROSCon. @VinDp will mainly work on this.

> Also this prs are using a different approach enabling free drive via ur script and not with the new function in the client library. Why do you close this active one and not the others?

I mainly closed this one because it was the third one and id didn't mention looking at the others. I've taken a closer look at #546, #573 and this one and indeed, this one seems to have the best approach. Sorry for not looking at this in more detail before.


That being said, I see one main concern about this approach:

Since this is using the freedrive support of the ur_driver, the reverse_interfaces is being kept active, so any other controller will stay active. From my understanding that would mean that the JTC will keep sending the last position of the latest trajectory all the time. Once freedrive mode is exited, the program running on the robot will return to reading the commands being sent on the reverse interface which will still be the old ones, I suppose.

@VinDp will test this on a real robot later this week to confirm / disprove my assumption.

I would much more love to have this into an own controller that can actually claim the command interfaces, so the JTC has to be deactivated and re-activated explicitly. We'll do that for ROS 2 as we don't have the option to advertise ROS interfaces from the hardware interface directly.

fmessmer:
thanks @fmauch @VinDp for getting back at this
@schmid-jn already used this significantly on one or our URs and - as far as I know - did not observe anything weird, but I haven't been directly involved
we are curious about @VinDp findings and proposal and are happy to test once something is available...

urfeex:

PR #721, opened 2024-09-09
Specify velocities for trajectory forwarding test

fmauch:
Otherwise the trajectory points will never be sent.

### DIFF
--- a/ur_robot_driver/test/integration_test.py
+++ b/ur_robot_driver/test/integration_test.py
@@ -320,6 +320,7 @@ class IntegrationTest(unittest.TestCase):
         for i, position in enumerate(position_list):
             point = JointTrajectoryPoint()
             point.positions = position
+            point.velocities = [0, 0, 0, 0, 0, 0]
             point.time_from_start = rospy.Duration(duration_list[i])
             goal.trajectory.points.append(point)
         for i, joint_name in enumerate(goal.trajectory.joint_names):


urrsk:
Why does it never get sent?
Due to our new check on splines?

VinDp (approved):

fmauch:
> Why does it never get sent? Due to our new check on splines?

Basically because of the logic in [`HardwareInterface::startJointInterpolation()`](https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/a8e2b0b7d80e769facc49d51e641baec72c4e5f1/ur_robot_driver/src/hardware_interface.cpp#L1265-L1325)

```c++
    if (!use_spline_interpolation_)
    {
      // write point for movej command
    }
    else  // Use spline interpolation
    {
      if (point.velocities.size() == 6 && point.accelerations.size() == 6)
      {
        // write quintic spline segment
      }
      else if (point.velocities.size() == 6)
      {
        // write cubic spline segment
      }
      else
      {
        // nothing being sent
        ROS_ERROR_THROTTLE(1, "Spline interpolation using positions only is not supported.");
      }
  }
```

In the test we use a position-only trajectory so we end up in the last case, where no trajectory point is being sent, as we default to spline-interpolated execution. Since in the past we used the `movej` interface by default, the test actually was working correctly and got missed when switching to spline interpolation by default, as it was falsely succeeding. https://github.com/UniversalRobots/Universal_Robots_Client_Library/pull/184 implements actually failing if no trajectory point was received, so the trajectory execution in the test fails.

urrsk:
That makes sense.
Then I will suggest that we add check and an error message to explain why the robot is not moving on the ROS side as well

urrsk (approved):

PR #724, opened 2024-09-20
Update branch of support repo for noetic

fmauch:
Since a while we moved to a `noetic-devel` branch on the support repo.


VinDp (approved):
Nice update :+1:

PR #767, opened 2026-02-17
Explicitly send MODE_STOPPED when returning control to the robot

urfeex:
This way, we can distinguish between an intentional hand-back and a communication loss

This is connected to https://github.com/UniversalRobots/Universal_Robots_Client_Library/issues/438


urrsk (changes requested):
Looks good. But please add a test covering this

urrsk (approved):

